PR #1, opened 2014-11-07
[NEW MODULE] Additions for eCommerce under German Law

mncoon:
This addon installed necessary additions in accordance with statutory requirements in Germany.
## Overview
- adds terms page
- adds revocation page
- adds delivery page
- adds privacy page
- adds imprint page
- adds an easy built-in address snippet to content blocks
- adds "incl. VAT, plus shipping costs" to every price - adds terms and revocation popup on payment site; both are taken from terms and revocation page on the frontend


OSguard:
Hi @ma78nee 

Do you need some help to interpret the problems from the travis output? You can click on [Details](https://travis-ci.org/OCA/l10n-germany/builds/40279681) and see what is wrong.

For my first review i suggest the following thinks:
- split the module in three smaller modules
  1. Adding new pages
  2. address snippet block
  3. "incl. Vat, plus shipping costs"
- adding new pages we don't need to write python code, you can do this as well only by using xml. Example you can find in the [Core Source](https://github.com/odoo/odoo/blob/8.0/addons/website/data/data.xml). The benefit of having a module without Python code is that we can deploy it to any SaaS Installation with the "base_import_module".

mncoon:
Hi @OSguard 

Thank you for your feedback. I try to do the changes in the next days. 
I close this request and make some new. Also there is the "l" missing at the beginning oft the module.

OSguard:
you can keep the pull request open, just push the changes to the branch witch you want to contribute, then the pull request will be updated. And we can add a label "work in progress".
But now it is to late. Make pull request for each separate module and then we are working on the improvement.

mncoon:
@OSguard 

What I found out in adding new pages without python is, that the url goes: `host.com/web/your_module_name.page_name`, as defined in the template tag of the view xml: `<template id="terms" name="Terms" page="True">`.

For example: `host.com/web/l10n_de_website_sale_accept_terms.terms`. That is not really practicable, neither for users, nor for search engines.

Even every core-website-module uses the @http.route to reroute for example `host.com/web/website_sale.products...` to `host.com/shop/product...`.

For the popups on the payment page I need the controller anyway, so I let the routes also there. I hope thats ok.

PR #3, opened 2014-12-11
Adds the states in Germany

ntodorova:
This module adds the states in Germany.


OSguard:
l10n_de_country_states/**init**.py:23: [W0403(relative-import), ] Relative import 'res_country', should be 'l10n-germany.l10n_de_country_states.res_country'

this module is 7.0? you push to 8.0 branch

PR #17, opened 2015-11-17
[ADD] l10n_de_steuernummer addon

antespi:
This module was written to extend the functionality of base_vat to support validation of german SteuerNummer following this schema: https://de.wikipedia.org/wiki/Steuernummer

By now, this validator only checks (after official DE validator), that VAT is 10 or 11 digit length


pedrobaeza:
You shouldn't overwrite the field vat for any other use, because this field has other implications along the modules. Create a new field for this purpose.
:-1:

rafaelbn:
Hi @OSguard , how do you validate SteuerNummer ? Thanks

antespi:
Sorry, I don't understand your point. Steuernummer is a legal VAT ID in German, it is the old version of current Steue-IdNr but completely legal. It is like CIF and NIF in Spain, but format are different. There are companies and persons in German that only have Steuernummer and need it for taxing

rafaelbn:
Rebuilding for testing

pedrobaeza:
The problem is to introduce the number without 'DE', so I still stand in my position. All VAT numbers should be homogeneous starting with the country code.

antespi:
@pedrobaeza This PR holds DE prefix. This approach only changes `check_vat_de`, so user must prefix steuernummer by DE in order to validate it. You can check it in runbot with this Steuenummer: `DE201/123/12340` (valid) and this without DE prefix `201/123/12340` (invalid)

pedrobaeza:
Please add a test for keeping coverage and I'll merge.

rafaelbn:
Tested in production enviroment of one customer :+1:

rafaelbn:
@JavierIniesta please review

tv-openbig:
@pedrobaeza Please don't merge that, before Rafael explains better.
For what tax case and purposes would you need the module you want to merge ?

The VAT-ID controls a feature as described in the base_vat module description should be the same in each european union country. We don't need to change that.  Probably you are totally misleaded by wrong terms for "name" and "description" in res.company and res.partner object. **It is an odoo issue** .
Nevertheless it can be bypassed easily, f.e. by the translations or simply by correct usage, however it remains in practise an error, which may be a barrier for non-accountants each time. 

Personally i have never created a bug so far, as i wasn't 100% sure, if the reason could have affected non EU countries, where that terms match better. With a little bit understanding of Odoo accounting it is very easy to configure Odoo in the way you need it. The terms mismatch is not a critical problem, however it is again and again misleading.

Currently we have  to separate from german point of view the two terms: 
1. **VATID *_(in german: Umsatzsteuer-ID) and 
2. *_Tax ID** (in german Steuernummer). 

Both are different things. 

# **Case 1.**VATID is used to get the VATID on the invoice document.

VATID is a legal requirement to have it black on white on the invoice document. That number is only used for the tax type VAT. 

To my opinion that field should be the field "vat" on res.partner object.
The current fields name is "TIN". 

**How it should be ?**
Fields name should remain "vat", however fields description should be changed 
from "TIN" to  "VAT".  That would be more clear.

# **Case 2.** The TAX ID is the tax number which you need to enter into  the field Tax ID

to declare various tax types, f.e. by your monthly electronical declaration. 
That number is needed for various tax types, and the government needs it 
f.e. to match payment entries with your declaration. The same number is used 
for all tax declarations of each tax type, f.e. to declare the tax of income, property tax 
and so on. VAT is only one of all possible tax type cases.

As we currently don't have a module to declare tax by electronical transfer, 
in any official Odoo module, we don't need that information (so far) in Odoo.
However if you want to develop a module to do that, you would indeed 
need to edit somewhere Tax ID to use it for your various monthly or per year 
tax declarations.

To my opinion the field to use for that is the field "vat" on res.company object.
The fields description is "Tax ID", from user point of view that would fit perfectly, 
from developer point of view the field name "vat" is a bit misleading, as i already 
mentioned.. 

**How it should be ?**
Fields name should be "taxid" and fields description should be "VAT".

antespi:
@tv-openbig Sorry, I can't understand your point. 

This is an addon (that means, it is optional). If you want Odoo standard don't install it. If you want to invoice to your german customers using old german Steuernummer then install it.

Please @copmn, @OSguard, @pedrobaeza give your opinion about this. For me there is no problem to move this addon to non-OCA github repo, but I think this is a good addon for OCA.

Thanks all

pedrobaeza:
I think @antespi is right and the purpose of this module is good justified. Anyway, I still ask for a little test to keep the coverage.

mncoon:
For me the addon is ok. Maybe it could be useful, because in germany there are small companies or other professions like doctors which don't have or can't get a EU-VAT-ID.

antespi:
Test is pending, I know. I'll add it as soon as possible

OSevangelist:
I am afraid after discussing the matter with @OSguard i do not agree. Reasoning is as follows. German tax law differentiates between VAT-ID and TAX-ID and in some cases you might need both numbers so i would highly recommend to extend the module instead of substituting apples with beans. If somebody understands German here is a good explanation why keeping both numbers is advised

https://debitoor.de/gruenderlounge/handel-europa/muss-die-umsatzsteuer-id-oder-die-steuernummer-auf-die-rechnung

mncoon:
@OSevangelist the link which you have posted is for the numbers of your own company. 

It is true that you can use both of them in invoice footer. But in company settings there is also only one field which you can use as you want with EU-VAT-ID or with Germany VAT-Number.

Please keep in mind that this is onyl additional and it will only check if the number is valid (= correct schema). It does nothing else. Everyone is free to use it or not. And everyone who want's to use it can freely decide what number he/she wan't to put in.

rafaelbn:
Agree with @copmn 

"VAT-ID and TAX-ID and in some cases you might need both numbers", when? Talking only about customers.
- If you are a German Company and you invoice to German Customers you will use 1 number.
- If you are a German Company and you invoice to UE customer you will use 1 number
- If you are a Spanish Company and you invoice to Spanish company you will use 1 number
- If you are a Spanish Company and you invoices a German company which doesn't have VAT-ID you will use 1 number: TAX-ID (Steuernummer)
- If you are a Spanish Company and you invoice a German company which have VAT-ID you will user 1 number: VAT-ID

IMHO for this PR is an optional one from some Companies which need it and is :+1: 

Thanks

OSevangelist:
@copmn @rafaelbn well I do understand your points but although I agree in most -if not al cases- one would use just one of them regularly, I still think that VAT-ID and TAXID are different things from a semantical perspective and should therefore be stored in different attributes.

pedrobaeza:
@OSevangelist, for only that unusual case that you need both numbers, it doesn't worth to modify all reports, views, and so on to adapt them for the new field, don't you think?

OSevangelist:
@pedrobaeza don't get me wrong but eventhough that means to change a lot of reports I would still avoid mixing apples with beans (i.e. TaxID with VAT-ID). However, I do agree that this might be viewed as a chicken and egg problem. In turn, what does it take to use an "if / else construct" in all reports? Isn't that merely an copy and paste effort and valuable just due to the fact that it keeps semantical correctness of the attribute? At the end of the day every user can decide whether he / she wants to use that quickfix so merge if you think the advantages outweigh the disadvantages and let's refocus on the next steps. Have a nice weekend

pedrobaeza:
OK, thanks for the understanding. @antespi, then please make 2 things:
- Make clear in the README the clarification made by @OSevangelist about both fields.
- Create a little test.

Have a nice weekend too!

rafaelbn (approved):
Please @antespi attend https://github.com/OCA/l10n-germany/pull/17#issuecomment-180405326 as commented. We must merge this before migrate it to v9. Thanks. cc @pedrobaeza @Tecnativa 👍

pedrobaeza:
Closing in favor of #23

PR #24, opened 2017-08-16
[FIX] Upgrade .travis.yml

chienandalu:
Upgrade travis according to quality tools.

cc @Tecnativa


pedrobaeza (approved):
Runbot error is expected as there's no addons for now. I fast-track

PR #25, opened 2017-08-25
10.0 base country state translatable

rami-wafaie:


pedrobaeza:
But this should be in OCA/partner-contact, not here...

rami-wafaie:
Will be closed and new PR to Partner Contact will be created

PR #31, opened 2018-02-02
11.0 mig l10n de steuernummer

rami-wafaie:


pedrobaeza (approved):
Please squash your last 2 commits

satriani-vai:
There are two vat numbers in Germany, StNr / Steuernummer (e.g. 012/123/12345) and the USt-IdNr / Umsatzsteuer-Identifikationsnummer (e.g. DE123456789). Both are very important.
So, I would ad a new field, extra for the StNr / Steuernummer.

pedrobaeza:
@satriani-vai this can't be in the current form, as this field is reused in a lot of places. Having a second field, that one won't be used in that places.

Merging this one as trivial migration.

PR #33, opened 2018-07-03
[ADD] l10n_de_zip

tschanzt:
adds l10n_de_zip which provides the data for res.better.zip so we can autocomplete locations in Germany.

### DIFF
--- a/l10n_de_zip/__init__.py
+++ b/l10n_de_zip/__init__.py
@@ -1,0 +1,2 @@
+# Copyright 2018 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
--- a/l10n_de_zip/__manifest__.py
+++ b/l10n_de_zip/__manifest__.py
@@ -1,0 +1,21 @@
+# Copyright 2018 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+{
+    'name': 'Germany - Postal codes (ZIP) list',
+    'summary': 'Provides all German postal codes for auto-completion',
+    'version': '11.0.1.0.0',
+    'category': 'Localisation/Europe',
+    'licence': 'AGPL-3',
+    'depends': [
+        'base',
+        'base_location',
+        'l10n_de_country_states'
+    ],
+    'author': 'Camptocamp, Odoo Community Association (OCA)',
+    'website': 'http://www.camptocamp.com',
+    'data': [
+        'data/l10n_de_zip.xml'
+    ],
+    'installable': True,
+}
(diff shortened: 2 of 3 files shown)


pedrobaeza:
A better module name would be `l10n_de_location`, following other conventions. Isn't Geonames providing these ZIP entries through https://github.com/OCA/partner-contact/tree/11.0/base_location_geonames_import?

I see 2 problems with current approach:

* They are static data to be maintained and will be obsolete very soon
* The module take a lot of time to be installed.

You can check the approach taken in other countries like Spain here: https://github.com/OCA/l10n-spain/tree/11.0/l10n_es_toponyms

OSevangelist:
@pedrobaeza @tschanzt i briefly reviewed the spanish approach and the data provided by the the plattform (seems accurate and detailed for DE as well) so i also strongly advocate to adopt the "standard way" of doing things as suggested by @pedrobaeza rather than having another module with potentially diverting data doing the same thing but only through different code

simahawk:
[ Disclaimer: not very familiar w/ the tool :wink: ]

@pedrobaeza @OSevangelist I perfectly agree but I'm a bit perplexed about which way is the standard one :smile: 

What I see here https://raw.githubusercontent.com/OCA/l10n-spain/11.0/l10n_es_toponyms/wizard/l10n_es_toponyms_zipcodes.xml is static file.

The only difference is the script to update it here https://github.com/OCA/l10n-spain/blob/11.0/l10n_es_toponyms/gen_src/gen_toponyms_geonames.py
which is using the same data as https://github.com/OCA/partner-contact/blob/11.0/base_location_geonames_import/wizard/geonames_import.py#L176

Am I wrong? 

Trying to summarize:
1. do not load everything at install
2. use the wizard not to fetch directly but to load the static file
3. provide a script to update the file (this one could be made generic maybe)

I guess that point 2 is just to be able to diff and to avoid download issues? Otherwise I don't see the real reason for it.

pedrobaeza:
@simahawk I'm afraid you are wrong. That are fallback data in case there's no online connection. The default one is to load from Geonames, and next version I'm going to remove that fallback, as Geonames has proven to not remove this info (when this was created, Geonames changed several times the access to these files).

Please double check.

simahawk:
good to know. @tschanzt pls update as required

tschanzt:
@pedrobaeza @OSevangelist @simahawk I'm closing the PR, since we already have an import for the zip codes and we don't need to customize something for Germany (at least not that I know of :smile: ) it doesn't make sense to have a special module for it.

pedrobaeza:
OK, great. We need it because the states are not mapped correctly due to the columns that set the regional partition are different, but if it's not your case. The module also serves for focusing people who wants to know how to do it. Maybe an empty module only depending on base_location_geonames_import and with instructions in German is a good idea...

PR #38, opened 2019-02-25
module with german vat tax declaration report

tv-openbig:
We have cloned and adopted the l10n-netherlands tax report module from
Onestein B.V.  The report is constructed in the same way as official
german „Umsatzsteuervoranmeldung“. The values from the report could be
easily transmitted via official portal https://www.elster.de/ in order
to declare and transfer your monthly (or quarterly) vat taxes.


tv-openbig:
I have just sent my CLA's to cla@odoo-community.org.

tv-openbig:
Hello,
@pedrobaeza @astirpe currently i do not know how to proceed. 

The pylint check is now finally ok:
Your code has been rated at 10.00/10
count_errors 0
Starting lint check only for modules changed

Runbot and also the two final checks against OCB and odoo repository failed because of missing module account_tax balance which is a dependency of the proposed module l10n_de_tax_statement (https://github.com/OCA/account-financial-reporting/tree/11.0/account_tax_balance).

Do i have to put it in the l10n-germany repository or is something else missing 
to pass the runbot test?

Here are the relevant lines, from odoo startup log :
2019-02-27 19:08:55,138 6298 WARNING openerp_template odoo.modules.loading: invalid module names, ignored: account_tax_balance

File "/home/travis/OCB-11.0/odoo/addons/base/module/module.py", line 363, in _state_update
raise UserError(_("You try to install module '%s' that depends on module '%s'.\nBut the latter module is not available in your system.") % (module.name, dep.name,))
odoo.exceptions.UserError: ("You try to install module 'l10n_de_tax_statement' that depends on module 'account_tax_balance'.\nBut the latter module is not available in your system.", '')

Thanks a lot in advance for your help.

pedrobaeza:
Create an oca_dependencies.txt file in root (if not exists), and put inside `account-financial-reporting`. With this, you are instructing Travis/Runbot for adding that OCA repository, and thus having accessible `account_tax_balance` module.

tv-openbig:
@pedrobaeza Thanks a lot. Now further checks went through. Obviously now I have to clean / correct the tests at first.

l10n_de_tax_statement/wizard/l10n_de_tax_statement_config_wizard.py
```diff
@@ -0,0 +1,232 @@
...
+            defv.setdefault('tag_62_tax', config.tag_62_tax.id)
+            defv.setdefault('tag_67_tax', config.tag_67_tax.id)
+            defv.setdefault('tag_63_tax', config.tag_63_tax.id)
+            defv.setdefault('tag_64_tax', config.tag_64_tax.id)
+            defv.setdefault('tag_59_tax', config.tag_59_tax.id)
+            defv.setdefault('tag_69_tax', config.tag_69_tax.id)
+            return defv
+
+            if (not self._is_l10n_de_coa_skr03() or not self._is_l10n_de_coa_skr04()):
+                return defv
```
fkantelberg: This part of code will never be reached. Is the indentation level right?

l10n_de_tax_statement/wizard/__init__.py
```diff
@@ -0,0 +1,5 @@
+# Copyright 2019 BIG-Consulting GmbH (<http://www.openbig.org>)
+# Copyright 2019 Onestein (<http://www.onestein.eu>)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from . import l10n_de_tax_statement_config_wizard
```
fkantelberg: A blank line at the end of the file is missing

l10n_de_tax_statement/__manifest__.py
```diff
@@ -0,0 +1,26 @@
...
+
+{
+    'name': 'German VAT Statement',
+    'version': '11.0.1.1.0',
+    'category': 'Localization',
+    'license': 'AGPL-3',
+    'author': 'OpenBIG.org, Onestein, Odoo Community Association (OCA)',
+    'website': 'https://github.com/openbig/l10n-germany',
+    'depends': [
+        'account_tax_balance',
```
fkantelberg, line 14: Missing dependency to the module l10n_de for the tags

l10n_de_toponyms/__manifest__.py
```diff
@@ -16,6 +16,7 @@
     "license": "AGPL-3",
     "data": [
         "wizard/l10n_de_toponyms_wizard.xml",
+        "wizard/l10n_de_toponyms_zipcodes.xml",
```
fkantelberg, line 19: As far as I know is this just an example file you can get using the `tools/gen_toponyms_geonames.py`. This shouldn't be included in the data. Furthermore this are changes in another module and shouldn't be in this PR to keep the PR clean.

l10n_de_tax_statement/models/l10n_de_tax_statement_config.py
```diff
@@ -0,0 +1,63 @@
...
+    tag_85_tax = fields.Many2one('account.account.tag')
+    tag_65_tax = fields.Many2one('account.account.tag')
+    tag_66_tax = fields.Many2one('account.account.tag')
+    tag_61_tax = fields.Many2one('account.account.tag')
+    tag_62_tax = fields.Many2one('account.account.tag')
+    tag_67_tax = fields.Many2one('account.account.tag')
+    tag_63_tax = fields.Many2one('account.account.tag')
+    tag_64_tax = fields.Many2one('account.account.tag')
+    tag_59_tax = fields.Many2one('account.account.tag')
+    tag_69_tax = fields.Many2one('account.account.tag')
```
fkantelberg: It would be a bit clearer and better for debugging if you can sort them.

l10n_de_tax_statement/tests/test_l10n_de_tax_statement.py
```diff
@@ -0,0 +1,356 @@
...
+    def test_08_update_exception2(self):
+        self.config.unlink()
+        with self.assertRaises(UserError):
+            self.statement_1.statement_update()
+
+    def test_09_update_working(self):
+        self.invoice_1._onchange_invoice_line_ids()
+        self.invoice_1.action_invoice_open()
+        self.statement_1.statement_update()
+        self.assertEqual(len(self.statement_1.line_ids.ids), 22)
```
fkantelberg: Aren't there 47 tags?

fkantelberg (commented):
Hello

I checked why the Travis CI are alerting and marked main problems. You should check the tests and adjust them to the german module.

The translations are mixed up. There are english and german phrases.

l10n_de_tax_statement/__manifest__.py
astirpe: I think that the dependency on module `l10n_de` should be not here. The reason is that this module was meant to be compatible with every German localization module (eg.: any custom localization module for Germany).
The logic implemented in the [configuration wizard](https://github.com/OCA/l10n-germany/blob/937ccd80bd174876e6f8534c99743aaa1dc6ff7e/l10n_de_tax_statement/wizard/l10n_de_tax_statement_config_wizard.py) is adaptive: if `l10n_de` is already installed, when installing this module the tags are recognized and included in the configuration.

l10n_de_tax_statement/wizard/l10n_de_tax_statement_config_wizard.py
astirpe: Yes, I think there is some indentation issue here.
@tv-openbig

tv-openbig: I would change that due to long line error to:
if (not self._is_l10n_de_coa_skr03() or
                    not self._is_l10n_de_coa_skr04()):
                return defv

I needed to change that (in comparition to netherland module l10n_nl_vat_statement) as we have with "l10n_de_skr03" or "l10n_de_skr04" two accounting plans as a choice for german companies. At first the pre-configured tax tag mapping to the tax statement positions was working just for SKR03 but not for SKR04. Now it works for both accounting plans. Indeed i am also not 100% sure if the lines in between 114 - 158 are required or not. Eventually they are usefull if none of above accounting charts would be selected (just l10n_de module with a sel-made accounting plan).

l10n_de_tax_statement/__manifest__.py
fkantelberg: In that way the adaptive implementation isn't working because the tests are throwing errors because of the missing xmlids. In this case you should check if the `l10n_de.tag_de_xx` are present in the lines 114+ in the wizard.

Which different german localization modules do you mean?

l10n_de_tax_statement/wizard/__init__.py
tv-openbig: Thanks. This morning i found these further lines in the reporting, which i hopefully have fixed alreaday:
======== Testing test_flake8 ========

l10n_de_tax_statement/wizard/l10n_de_tax_statement_config_wizard.py:111:80: E501 line too long (86 > 79 characters)

l10n_de_tax_statement/wizard/l10n_de_tax_statement_config_wizard.py:159:29: E211 whitespace before '('

l10n_de_tax_statement/wizard/l10n_de_tax_statement_config_wizard.py:165:29: E211 whitespace before '('

l10n_de_tax_statement/tests/test_l10n_de_tax_statement.py:6:1: F401 'mock.patch' imported but unused

l10n_de_tax_statement/tests/test_l10n_de_tax_statement.py:12:1: F401 'odoo.tests.common.post_install' imported but unused

l10n_de_tax_statement/tests/test_l10n_de_tax_statement.py:12:1: F401 'odoo.tests.common.at_install' imported but unused

l10n_de_tax_statement/tests/test_l10n_de_tax_statement.py:356:1: W391 blank line at end of file

l10n_de_tax_statement/__manifest__.py
tv-openbig: Thanks, indeed i was thinking about that too, because i in the adopted tests i have referenced this tag. Hopefully that will fix the testing...

astirpe: @fkantelberg This module was initially made for the Netherlands localization. In the Netherlands the Chart of Accounts (COA) is free, every company could adopt its own COA, not necessarily the standard one provided by Odoo SA. Not sure is the same case for Germany.

l10n_de_toponyms/__manifest__.py
tv-openbig: In fact i stumbled over it as i comitted l10n_de_tax_statement as error notifications. Obviously these issues didn't happen as the original authors committed that module.  Actually i don't know if errors belonging to that module persists.

l10n_de_tax_statement/__manifest__.py
fkantelberg: Ah. Can we generalize this module and offer a glue module if the l10n_de is installed with the wizard configured properly?

l10n_de_tax_statement/wizard/l10n_de_tax_statement_config_wizard.py
astirpe: @tv-openbig the problem is the indentation. As it is right now, the two lines are indented "too much".

l10n_de_tax_statement/models/l10n_de_tax_statement_config.py
tv-openbig: On a first view it looks like a chaos ordering, however this ordering reflects the ordering of the official VAT declararation, but in fact you are right for debugging other ordering could make more sense for technical people.

l10n_de_tax_statement/tests/test_l10n_de_tax_statement.py
tv-openbig: yes you are right, 22 comes from the NL module.

l10n_de_tax_statement/__manifest__.py
astirpe: @fkantelberg when this logic will be fixed and working properly, not any glue module will be necessary (that was exactly the reason of implementing this logic).

l10n_de_tax_statement/__manifest__.py
```diff
@@ -11,6 +11,7 @@
     'website': 'https://github.com/openbig/l10n-germany',
     'depends': [
         'account_tax_balance',
+        'l10n_de',
```
astirpe: So at the end the dependency on `l10n_de` was added :smile: 
If you make this module dependent on `l10n_de`, the configuration mechanism can be simplified (especially the wizard). On the other side, by doing this way, you are forced to install `l10n_de` even if you don't use it.

l10n_de_tax_statement/wizard/l10n_de_tax_statement_config_wizard.py
astirpe: @tv-openbig still not ok. You have to shift those 3 lines of code to the left.

l10n_de_tax_statement/__manifest__.py
fkantelberg: This is why I suggested a glue module (dependency to l10n_de and l10n_de_tax_statement with auto_install = True). It will get automatically installed if both of the dependency is fulfilled. In that case you don't have to add any checks because Odoo is doing it for you.

astirpe: @fkantelberg did you read my https://github.com/OCA/l10n-germany/pull/38#discussion_r263787940 ?

l10n_de_tax_statement/wizard/l10n_de_tax_statement_config_wizard.py
tv-openbig: like this ?

            defv.setdefault('tag_59_tax', config.tag_59_tax.id)
            defv.setdefault('tag_69_tax', config.tag_69_tax.id)
            return defv

        if (not self._is_l10n_de_coa_skr03() or
                not self._is_l10n_de_coa_skr04()):
            return defv

        defv.setdefault('tag_41_base', self.env.ref('l10n_de.tag_de_41').id)
        defv.setdefault('tag_44_base', self.env.ref('l10n_de.tag_de_44').id)

l10n_de_tax_statement/__manifest__.py
tv-openbig: On my local test system i could install the module and it was working fine for both scenarios (without the l10n_de dependency in the module __manifest__.py):
a. German company with accounting plan skr03.
b. German company with accounting plan skr04.

Anyway finally a german company will ending either with l10n_de + l10n_de_skr03 or l10n_de + l10n_de_skr04 if they will decide to add on top this tax declaration module (l10n_de_tax_statement).

Theoretically there could be a scenario "c." : a german company doesn't want to install l10n_de and / or l10n_de_skr03 or l10n_de_skr04 neither. 

@astirpe I assume that is the scenario you mean ?

l10n_de_tax_statement/wizard/l10n_de_tax_statement_config_wizard.py
astirpe: Yes

l10n_de_tax_statement/__manifest__.py
fkantelberg: Yes but I don't see any reason to implement tests which are already in odoo if you use the features like autoinstall in modules. But I just want to help to make the module more readable and usable.

l10n_de_tax_statement/__manifest__.py
astirpe: @tv-openbig Yes, is theoretical scenario "c" possible in a real case? If yes, then my previous comments would be still valid.

astirpe: So the initial idea for this module was to make it independent by the chosen localization module. But in case you already had the `l10n_de` installed (that would be often the case), this module can load the proper tags in the configuration automatically.

fkantelberg: @astripe Then I think it was a misunderstanding but to make this module independent you should just delete everything which depends on l10n_de and put the dependent stuff into a new glue module  which will automatically installed when l10n_de and this one are installed to configure for the l10n_de case. This would mean that the glue module won't be installed in the scenario c but in a and b where it can set the configuration for the german localization.

AStripe: > @astirpe I think this is for you. poor choice of user name on my part I guess.

astirpe: @fkantelberg
When I say "independent" I mean: not containing `l10n_de` in the dependencies of the `__manifest__.py` file.
When I say "adaptive" I mean: the configuration is automatically set in case `l10n_de` is already installed, which is a common case.

The glue logic would be fine for scenario "c". But when we talk about the standard Odoo localization modules (scenarios "a" and "b"), the glue logic can be avoided IMO. So the end users would be happy to just install one module instead of two or more.

I'm not very into the German community, so for me doesn't really matter which logic you choose. I just wanted to help to understand this module better. If you are happy with two modules instead of one for the case "standard German localization + OCA tax statement", I'm not going to block it :wink:

fkantelberg: @astirpe There was the misunderstanding. :+1: 
You want to minimize the number of module because the user only have to install 1. Same here. You can achieve it by setting the `auto_install` to true in the manifest. Odoo will install it if the dependencies are fulfilled.

astirpe: @fkantelberg Some customer could forget to download the auto_install modules, or just could not notice that those modules exist. And there are many other situations where you would prefer just one plug-and-play module, instead of handling multiple simple modules.
If you are worried about the code readability, still is possible to write a good code without splitting it in multiple modules.
But I'm not going to insist anymore. :smile:

tv-openbig: @astirpe @fkantelberg The module works fine. It could be installed on top of the two only existing modules without any issues. For the case a user wants to install that module without any l10n_de module installed i assume it would also work if you configure  the tax statement in advance by hand before you would be able to get a proper report. 

As a functional consultant i can basically understand the code. It was enough to adopt it. For sure it is more important to have background knowledge of german accounts, taxes, tax declaration as coding knowledge for that kind of module. For financial accountants as end users it is self-explainable.

The mixing of german terms and english terms is non-avoidable or better it doesn't make sense to back-translate the german terms in english.

I would be glad if we could push somehow that module through the testing as i wanted to proceed to create next PR's for german balance sheets / p&l report. If you could help to push that module through the remaining technical barriers it would be really great.

l10n_de_tax_statement/__manifest__.py
tv-openbig: @astirpe @fkantelberg I will now test locally what happens, if i remove the l10n_de dependency in order to simulate the scenario c.) i mentioned above, as for sure that could theoretically a user descicion, even if it is near 0% that it really would happen.
If that will work as expected i will remove the l10n_de dependency again.

l10n_de_tax_statement_zm/__manifest__.py
```diff
@@ -0,0 +1,22 @@
+# Copyright 2019 BIG-Consulting GmbH (<http://www.openbig.org>)
+# Copyright 2019 Onestein (<http://www.onestein.eu>)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+{
+    'name': 'German VAT Statement Extension',
```
astirpe, line 6: This module should have a more explanatory name, eg.: `German Intra-Community transactions declaration`.
You can also add a ["summary"](https://github.com/OCA/maintainer-tools/blob/master/template/module/__manifest__.py#L6) line, to add extra information, eg.: `Zusammenfassende Meldung` or whatever.

l10n_de_tax_statement/__manifest__.py
```diff
@@ -0,0 +1,27 @@
...
+# Copyright 2019 BIG-Consulting GmbH(<http://www.openbig.org>)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+{
+    'name': 'German VAT Statement',
+    'version': '11.0.1.1.0',
+    'category': 'Localization',
+    'license': 'AGPL-3',
+    'author': 'OpenBIG.org, Onestein, Odoo Community Association (OCA)',
+    'website': 'https://github.com/openbig/l10n-germany',
```
astirpe: `https://github.com/OCA/l10n-germany`

l10n_de_tax_statement/__manifest__.py
```diff
@@ -0,0 +1,27 @@
+# Copyright 2019 Onestein (<https://www.onestein.eu>)
+# Copyright 2019 BIG-Consulting GmbH(<http://www.openbig.org>)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+{
+    'name': 'German VAT Statement',
+    'version': '11.0.1.1.0',
```
astirpe: `11.0.1.0.0`

l10n_de_tax_statement/__manifest__.py
tv-openbig: @astirpe @fkantelberg I have now checked locally. 

If you select germany as country on database installation and install at first the dependency of the module l10n_de_tax_statement (=account_tax_balance) as a first step nothing happens as the dependency of account_tax_balance is imho wrong or outdated. Nothing happens means, i cannot see accounting menu, which is definitely a requirement if somebody wants to use the tax declaration module or wants to proceed testing on runbot.

I solved that so far by hand via installation of account_invoicing module. As a consequence of account_tax_balance dependency issue i would eventually propose to add that module as another dependecy in our module until that is fixed .

If in fact both comes together 1.) german company and 2.) account_invoicing module the l10n_de module as well l10n_de_skr03 and l10n_de_skr04 will be installed automatically. Now a user can  switch between both accounting plans until he creates first account moves. As a consequence for our l10n_de_tax_statement module the tax reporting could be used out of the box in both cases.

If i want to force scenario c.) described as „a german company wants to install some other non l10n_de related account chart“  i have to uninstall at first all official l10n_de modules if i marked the companies country as germany. If i install now f.e. l10n_nl or any other non l10n_de account chart indeed i get exactlly the error mentioned by travis automated testing:
  File "/odoo/odoo-server/odoo/addons/base/ir/ir_model.py", line 1324, in xmlid_lookup
    raise ValueError('External ID not found in the system: %s' % xmlid)
ValueError: External ID not found in the system: l10n_de.tag_de_41

I didn’t get that error anymore if l10n_de is a requirement of our module, even if i really want to use l10n_nl account chart in combination with l10n_de_tax_statement. 

Therefore i would propose to follow the proposal of @fkantelberg to add the l10n_de dependency.

The only drawback in that theoretical case would be that the tax statement configuration would need a reconfiguration to a.) correct the mappings to the NL tax tags or to adopt the NL tax -> DE tax tag mapping. I assume there is no other way to do that anyway. 

A future improvement could be that in a case if there is another accounting plan installed which doesn’t relate to  l10n_de to clean up the tax statement configuration wizard in order to allow a by hand tax statement configuration from scratch without any pre-filled tax statement configuration. I assume exactly that is covered by l10n_nl_vat_statement module in order to cover all theoretical cases of a account chart installation. 

Finally my proposal is to put that in the # TO DO section of the modules description, except anyone could help to solve that 5% case. Not to forget a reconfiguration could be done, so this  limited functionality is not an error. I assume if everybody agrees we should identify the lines we may uncomment in the source code. 

@astirpe Could you point to the lines to uncomment under that circumstances in l10n_de_tax_statement_config_wizard ?

astirpe: @tv-openbig 

The following error:

```
File "/odoo/odoo-server/odoo/addons/base/ir/ir_model.py", line 1324, in xmlid_lookup
raise ValueError('External ID not found in the system: %s' % xmlid)
ValueError: External ID not found in the system: l10n_de.tag_de_41
```
is caused by the wrong indentation issue, already mentioned in our previous comments.

Once you fix it, the dependency `l10n_de` could be safely removed.

l10n_de_tax_statement/views/l10n_de_tax_statement_view.xml
```diff
@@ -0,0 +1,140 @@
...
+                        <button name="finalize" string="Finalize" states="posted" type="object" class="oe_stat_button" icon="fa-stop-circle-o text-success" confirm="If you confirm, it will be not possible to modify this Statement or reset it back to draft anymore. Do you confirm?"/>
+                    </div>
+                    <label for="name"/>
+                    <h1>
+                        <field name="name"/>
+                    </h1>
+                    <group>
+                        <group name="tax_report">
+                            <field name="company_id" options="{'no_create': True}" groups="base.group_multi_company"/>
+                            <field name="date_range_id" attrs="{'readonly': [('state','=','posted')]}"/>
```
astirpe: You should also make the fields readonly when state is final.
See lines https://github.com/OCA/l10n-netherlands/blob/11.0/l10n_nl_tax_statement/views/l10n_nl_vat_statement_view.xml#L27-L29

l10n_de_tax_statement/README.rst
```diff
@@ -0,0 +1,73 @@
...
+   :alt: Odoo Community Association
+   :target: https://odoo-community.org
+
+This module is not yet maintained by the OCA.
+
+OCA, or the Odoo Community Association, is a nonprofit organization whose
+mission is to support the collaborative development of Odoo features and
+promote its widespread use.
+
+To contribute to this module, please visit https://odoo-community.org or the corresponding github repository.
```
astirpe, line 73: Could you please add the readme folder, following the template https://github.com/OCA/maintainer-tools/tree/master/template/module/readme, ad copy the content of this `README.rst` into fragment files in the readme folder?

See https://github.com/OCA/l10n-netherlands/tree/11.0/l10n_nl_tax_statement/readme

astirpe (changes requested):
I'm reviewing module `l10n_de_tax_statement` only (not `l10n_de_tax_statement_zm` yet).

l10n_de_tax_statement/__manifest__.py
astirpe: The indentation issue is causing this travis error:

```
l10n_de_tax_statement/wizard/l10n_de_tax_statement_config_wizard.py:111: [W0101(unreachable), VatStatementConfigWizard.default_get] Unreachable code
```

l10n_de_tax_statement/models/l10n_de_tax_statement_line.py
```diff
@@ -0,0 +1,174 @@
...
+        })
+        AccountTax = self.env['account.tax'].with_context(ctx)
+        return AccountTax.get_move_lines_domain(tax_or_base=tax_or_base)
+
+    def _get_domain_posted(self, taxes, tax_or_base):
+        self.ensure_one()
+        statement = self.statement_id
+        domain = [('move_id.l10n_de_tax_statement_id', '=', statement.id)]
+        if tax_or_base == 'tax':
+            tax_domain = [('tax_line_id', '=', taxes.ids)]
```
astirpe:
```suggestion
            tax_domain = [('tax_line_id', 'in', taxes.ids)]
```

l10n_de_tax_statement/models/l10n_de_tax_statement_line.py
```diff
@@ -0,0 +1,174 @@
...
+        else:
+            tax_domain = [('tax_ids', '=', taxes.ids)]
```
astirpe:
```suggestion
            tax_domain = [('tax_ids', 'in', taxes.ids)]
```

l10n_de_tax_statement/models/l10n_de_tax_statement_line.py
```diff
@@ -0,0 +1,174 @@
...
+                line.is_readonly = True
+
+    @api.multi
+    def unlink(self):
+        for line in self:
+            if line.statement_id.state == 'posted':
+                raise UserError(
+                    _('You cannot delete lines of a posted statement! '
+                      'Reset the statement to draft first.'))
+        super(VatStatementLine, self).unlink()
```
astirpe, line 113:
```suggestion
            if line.statement_id.state == 'final':
                raise UserError(
                    _('You cannot delete lines of a statement set as final!'))
        super(VatStatementLine, self).unlink()
```

l10n_de_tax_statement/wizard/l10n_de_tax_statement_config_wizard.py
```diff
@@ -0,0 +1,232 @@
...
+            defv.setdefault('tag_61_tax', config.tag_61_tax.id)
+            defv.setdefault('tag_62_tax', config.tag_62_tax.id)
+            defv.setdefault('tag_67_tax', config.tag_67_tax.id)
+            defv.setdefault('tag_63_tax', config.tag_63_tax.id)
+            defv.setdefault('tag_64_tax', config.tag_64_tax.id)
+            defv.setdefault('tag_59_tax', config.tag_59_tax.id)
+            defv.setdefault('tag_69_tax', config.tag_69_tax.id)
+            return defv
+
+            if (not self._is_l10n_de_coa_skr03() or not self._is_l10n_de_coa_skr04()):
```
astirpe:
```suggestion
        if not self.is_l10n_de_coa_skr03() or not self.is_l10n_de_coa_skr04():
```

l10n_de_tax_statement/wizard/l10n_de_tax_statement_config_wizard.py
astirpe:
```suggestion
            return defv
```

l10n_de_tax_statement/__manifest__.py
```diff
@@ -0,0 +1,28 @@
...
+    'category': 'Localization',
+    'license': 'AGPL-3',
+    'author': 'OpenBIG.org, Onestein, Odoo Community Association (OCA)',
+    'website': 'https://github.com/openbig/l10n-germany',
+    'depends': [
+        'account_invoicing',
+        'account_tax_balance',
+        'l10n_de',
```
astirpe:
```suggestion
```

l10n_de_tax_statement_zm/tests/test_l10n_de_tax_statement_zm.py
```diff
@@ -0,0 +1,139 @@
...
+from odoo.addons.l10n_de_tax_statement.tests.test_l10n_de_tax_statement\
+    import TestVatStatement
+
+
+class TestTaxStatementZM(TestVatStatement):
+
+    def _prepare_zm_invoice(self):
+        for invoice_line in self.invoice_1.invoice_line_ids:
+            for tax_line in invoice_line.invoice_line_tax_ids:
+                tax_line.tag_ids = self.tag_41_base
```
astirpe:
```suggestion
                tax_line.tag_ids = self.tag_3
```

astirpe (changes requested):
@tv-openbig please apply all the suggested changes. Hopefully travis tests will pass.

l10n_de_tax_statement/tests/test_l10n_de_tax_statement.py
```diff
@@ -0,0 +1,354 @@
...
+            limit=1)
+        self.assertFalse(config)
+
+        wizard.execute()
+
+        config = self.env['l10n.de.tax.statement.config'].search(
+            [('company_id', '=', self.env.user.company_id.id)],
+            limit=1)
+        self.assertTrue(config)
+        self.assertEqual(config.tag_86_base, self.tag_1)
```
astirpe:
```suggestion
        self.assertEqual(config.tag_81_base, self.tag_1)
```

l10n_de_tax_statement/tests/test_l10n_de_tax_statement.py
```diff
@@ -0,0 +1,354 @@
...
+        self.assertEqual(config.tag_81_base, self.tag_2)
```
astirpe:
```suggestion
        self.assertEqual(config.tag_86_base, self.tag_2)
```

l10n_de_tax_statement/tests/test_l10n_de_tax_statement.py
```diff
@@ -0,0 +1,354 @@
...
+        self.assertEqual(len(_27), 1)
+
+        self.assertFalse(_25.format_base)
+        self.assertFalse(_25.format_tax)
+        self.assertTrue(_25.is_group)
+        self.assertTrue(_25.is_readonly)
+
+        self.assertEqual(_26.format_base, '100.00')
+        self.assertFalse(_26.is_group)
+        self.assertTrue(_26.is_readonly)
```
astirpe:
```suggestion
        self.assertFalse(_26.is_readonly)
```

l10n_de_tax_statement/tests/test_l10n_de_tax_statement.py
```diff
@@ -0,0 +1,354 @@
...
+
+        self.assertEqual(_27.format_base, '50.00')
+        self.assertFalse(_27.is_group)
+        self.assertTrue(_27.is_readonly)
```
astirpe:
```suggestion
        self.assertFalse(_27.is_readonly)
```

l10n_de_tax_statement/__manifest__.py
tv-openbig: I have applied your proposed change. Thanks a lot for your intervention!

l10n_de_tax_statement/__manifest__.py
tv-openbig: I have applied your proposed change. Thanks a lot for your intervention!

l10n_de_tax_statement/views/l10n_de_tax_statement_view.xml
tv-openbig: I have applied your proposed change. Thanks a lot for your intervention!

l10n_de_tax_statement/README.rst
tv-openbig: I have applied your proposed change. Thanks a lot for your intervention!

l10n_de_tax_statement/models/l10n_de_tax_statement_line.py
tv-openbig: I have applied your proposed change. Thanks a lot for your intervention!

l10n_de_tax_statement/models/l10n_de_tax_statement_line.py
tv-openbig: I have applied your proposed change. Thanks a lot for your intervention!

l10n_de_tax_statement/models/l10n_de_tax_statement_line.py
tv-openbig: I have applied your proposed change. Thanks a lot for your intervention!

l10n_de_tax_statement/wizard/l10n_de_tax_statement_config_wizard.py
tv-openbig: I have applied your proposed change. Thanks a lot for your intervention!

l10n_de_tax_statement/wizard/l10n_de_tax_statement_config_wizard.py
tv-openbig: I have applied your proposed change. Thanks a lot for your intervention!

l10n_de_tax_statement/__manifest__.py
tv-openbig: I have applied your proposed change. Thanks a lot for your intervention!

l10n_de_tax_statement_zm/tests/test_l10n_de_tax_statement_zm.py
tv-openbig: I have applied your proposed change. Thanks a lot for your intervention!

l10n_de_tax_statement/tests/test_l10n_de_tax_statement.py
tv-openbig: I have applied your proposed change. Thanks a lot for your intervention!

l10n_de_tax_statement/tests/test_l10n_de_tax_statement.py
tv-openbig: I have applied your proposed change. Thanks a lot for your intervention!

l10n_de_tax_statement/tests/test_l10n_de_tax_statement.py
tv-openbig: I have applied your proposed change. Thanks a lot for your intervention!

l10n_de_tax_statement/tests/test_l10n_de_tax_statement.py
tv-openbig: I have applied your proposed change. Thanks a lot for your intervention!

l10n_de_tax_statement/wizard/l10n_de_tax_statement_config_wizard.py
```diff
@@ -0,0 +1,232 @@
...
+            defv.setdefault('tag_61_tax', config.tag_61_tax.id)
+            defv.setdefault('tag_62_tax', config.tag_62_tax.id)
+            defv.setdefault('tag_67_tax', config.tag_67_tax.id)
+            defv.setdefault('tag_63_tax', config.tag_63_tax.id)
+            defv.setdefault('tag_64_tax', config.tag_64_tax.id)
+            defv.setdefault('tag_59_tax', config.tag_59_tax.id)
+            defv.setdefault('tag_69_tax', config.tag_69_tax.id)
+            return defv
+
+        if not self.is_l10n_de_coa_skr03() or not self.is_l10n_de_coa_skr04():
```
tv-openbig: @astirpe I am afraid, this line is still not OK as we now have no mappings inside of the tax statement configuration, even if i install a german account chart. I have locally checked that the following detects the mapping for one chart but unfortunately not for the other. For sure i could separate the modules into l10n_de_tax_statement_skr03 and l10n_de_tax_statement_skr04, however i can imagine that it is somehow possible to use the  "if / or" clause.

The following change would work with the l10n_de_skr03 account chart. The same would work if we would replace skr03 -> skr04 either for the one or for the other account chart. Currently it detects no mapping.

line 111:        if not self.is_l10n_de_coa_skr03(): #  or not self.is_l10n_de_coa_skr04():
...
...
and uncommenting of the lines 165-169:
    def is_l10n_de_coa_skr04(self):
       is_l10n_de_coa_skr04 = self.env.ref(
            'l10n_de_skr04.l10n_chart_de_skr04', False)
        company_coa = self.env.user.company_id.chart_template_id
        return company_coa == is_l10n_de_coa_skr04

astirpe:
```suggestion
        if not (self.is_l10n_de_coa_skr03() or self.is_l10n_de_coa_skr04()):
```

Could you check with this?

tv-openbig: @astirpe Now it is perfect! All cases are working (both german account charts + configurable account chart). Thanks a lot for your intervention!!

astirpe (approved):
Technical review OK.

tv-openbig:
@pedrobaeza Is it now possible to merge this PR ?

pedrobaeza:
OK, as @astirpe has given his bless and no more reviewers are expected, let's merge it.

PR #41, opened 2019-04-10
Require oca/mis-builder

pniederlag:
require oca/mis-builder for updated/new skr3/skr4 accounting


pniederlag:
obosolete, will target branch 12 with new PR

pedrobaeza:
But why providing this in a separate PR. If you need it for any new module, include this change in the PR with that module.

PR #42, opened 2019-04-10
add oca_dependencies.txt

pniederlag:
add required dependencies from oca


pedrobaeza:
But why providing this in a separate PR? If you need it for any new module, include this change in the PR with that module.

pniederlag:
@pedrobaeza you are definitly right, just wanted to help @tv-openbig to get his PR in a green state. Will work with him to update his PR and abandon this one.

PR #43, opened 2019-04-11
[11.0][FIX] l10n_de_tax_statement: name of menu "German Tax Tags"

astirpe:
This PR fixes the name of menu "German Tax Tags" in module `l10n_de_tax_statement`, according to what is already written in the README file.
It also contains some minor code review.


tbaden (approved):
code review: lgtm

l10n_de_tax_statement/models/l10n_de_tax_statement.py
```diff
@@ -67,7 +68,7 @@ class VatStatement(models.Model):
     )
     format_tax_total = fields.Char(
         compute='_compute_amount_format_tax_total',
-        string='Verbl. Ust.-Vorauszahlung (66 - 67)'
+        string='Verbl. Ust.-Vorauszahlung'
```
benbrich, line 71: I think 'Open advance vat payment' could be a good term to get the german out of the code :-) and (66-67) is a good hint for the user, and IMO should be in the help instead of being deleted

l10n_de_tax_statement/wizard/l10n_de_tax_statement_config_wizard.xml
```diff
@@ -248,7 +248,7 @@
     </record>
 
     <record id="action_account_tax_statement_de_config" model="ir.actions.act_window">
-        <field name="name">Ust-Voranmeldung</field>
+        <field name="name">German Tax Tags</field>
```
benbrich, line 251: IMO 'periodic vat returns' would be the better fit here.

benbrich (approved):
Thanks for the work, LGT! Just two linguistics recommendations.

l10n_de_tax_statement/models/l10n_de_tax_statement.py
astirpe: @benbrich field '`format_tax_total`' is a technical field and its description is never displayed in any view. We can change the description and add an help to `format_tax_total`, but there will be not any visible result. Instead if we change the description of field '`tax_total`', that one would be visible to the end user. What do you think? Shall we change that one?

l10n_de_tax_statement/wizard/l10n_de_tax_statement_config_wizard.xml
astirpe: I think that the "German" string should be specified there because there are some other similar modules that do the same for other localizations, for example the Netherlands Vat Statement module.

l10n_de_tax_statement/models/l10n_de_tax_statement.py
benbrich: Thanks for the hint, could have checked :-) This sounds good to me.

l10n_de_tax_statement/wizard/l10n_de_tax_statement_config_wizard.xml
benbrich: I am with you on this one. Just not so sure with Tax Tags. Ist this the right expression do you know? The big blue erp 'brother' is using 'periodic vat returns'

astirpe: "Tax Tags" is the Odoo way :smile: 
This is just a configuration menu item, so I don't think 'periodic vat returns' is appropriate. Anyway I don't have any particular problem in changing it. Maybe we can wait and see what other community members think about it.

benbrich: LOL, as I said I not sure about the naming as well. Maybe @tv-openbig has an good idea before merging? ;-)

l10n_de_tax_statement/models/l10n_de_tax_statement.py
tv-openbig: @astirpe @benbrich To my opinion "Verbl. Ust.-Vorauszahlung (66 - 67)" is visible in a view (at least in V10 it was). In fact i do not really understand why that was touched, because it is helpful in the context of that view.
Auswahl_009
However that would't be a big issue, at least for myself i wouldn't need that information.  Regarding the description, i don't understand why "german" should be deleted, but anyway each of those changes are really of minor importance,

astirpe: The field [tax_total](https://github.com/OCA/l10n-germany/blob/f1cb35bfc943352083424796fd426726168d1eae/l10n_de_tax_statement/models/l10n_de_tax_statement.py#L65-L68) has description `string='Verbl. Ust.-Vorauszahlung (66 - 67)'`. I didn't change it.

The field [format_tax_total](https://github.com/OCA/l10n-germany/blob/f1cb35bfc943352083424796fd426726168d1eae/l10n_de_tax_statement/models/l10n_de_tax_statement.py#L69-L72) is a technical field and its description is never displayed. I changed the description of this one to avoid a warning in the log of Odoo appearing when installing the module.

I hope it's more clear now.

l10n_de_tax_statement/wizard/l10n_de_tax_statement_config_wizard.xml
astirpe: @benbrich can we keep "German Tax Tags" and move forward?

benbrich: Yes for sure.

astirpe:
@pedrobaeza would you merge this one?

PR #44, opened 2019-04-12
Add module l10n_de_account_fiscal_position v11

ntodorova:
This module is used to solve the problem of fiscal position mapping.
In the case where a product from type service and product from type consumable with the same tax have to mapped to different taxes.


pedrobaeza:
We have this exact case in Spain, but it's not handled this way, which can be dangerous, as product type serves for stock, not for account. What we do is to have 2 taxes, one for services, and another for stockable products, and we map each one to the corresponding taxes. If not, this would convert also taxes reports in a nightmare.

ntodorova:
Hi @pedrobaeza, yes we have discussed this option too. We have decided that we will try them both.
Do you mean that the taxes report for service products and consumable products have to be different?
I thought that service and consumable products have to be under the same tax in the tax reports and this solution (the module here) will make that possible. 
I though there is only one special case where we need different tax and account mapping for service and consumable products when we use Fiscal Positions.  
But may be @tv-openbig knows better how this should be from accounting point of view.

pedrobaeza:
In Spain, some of them goes to different report fields, so doing through fiscal positions limits to only have the taxes mapping when you have a target tax, so in the end, you would need the same different target taxes. Why not having also different source taxes that map each one to a different target tax?

Example in Spain:

- Tax A: 21% VAT (Services)
- Tax B: 21% VAT (Goods)

On fiscal taxes report, there's a field for 21% VAT, so we sum A + B, but there are 2 fields for intra-community taxes, and thus we have a fiscal position called "Intra-community regime", with the mappings:

- Tax A > Tax C: intra-community 21% VAT (services)
- Tax B > Tax D: intra-community 21% VAT (goods)

and redirect C and D to the proper fields. If tomorrow, Spanish Fiscal Authority decides to also split national VAT in 2 fields, we only have to change the mapping, not the data.

ntodorova:
Sounds reasonable, but since I have a limited knowledge in the German accounting, I can not decide if this will also fit the German Requirements, so I will wait a bit till someone with more experience in the area enters the discussion.
Thank you @pedrobaeza!

tv-openbig:
@ntodorova Sorry Nikolina for my late reply. 
@pedrobaeza Indeed both options should normally fix the report issue (2 lines in the report instead of 2 columns). However there are some good arguments to accept Nikolina's PR :
No further duplication of local 7% / 19% tax (we already have tax included / tax excluded for the B2B / B2C case) in order to not confuse the average accounting users.
The default "national" tax assigned to a new product might be wrong (question if the property should should cover a "physical product" or a "service product", except a modified property assignation depends of the product type). 

Still i also see good chances that an accountant will be able to read from the current report the intra-community values out of two columns for physical and service products in order to enter two lines into the official formular.

pedrobaeza:
Thanks for your opinion. I let you decide then. Any way, if you finally go this way, module can be localization agnostic, and be put in https://github.com/OCA/account-fiscal-rule

tv-openbig:
@pedrobaeza Exactly the module is currently not only related to l10n-germany. However it would be nice to add the relevant tax / account mappings for either the product type "services" or "consumables" to the existing german fiscal position. The remaining product type should be selected in the empty field of the origin fiscal position line.  Probably that could be done in a separated "data" module or under the "data" directory of the "l10n_de_tax_statement_zm" module. Without adoption of the fiscal position i assume configuration would be really advanced topic for the average accounting user.

pedrobaeza:
@tv-openbig that's why we prefer to not follow this approach, as we can't bundle full CoA in standard `l10n_es`, but required extra modules, and thus, being exposed to not have complete chart in the company on first install.

There's also a problem with the module in current form: it doesn't take into account `product` type products (got with stock module), and it forces you to create 2 FPs, one for consumable and another for stockable products.

PR #45, opened 2019-04-16
balance sheet and profit & loss report for l10n_de_skr04

tv-openbig:
The module provides balance sheet and profit & loss report for german
account chart SKR04 (l10n_de_skr04). It is based on mis_builder module.


tbaden (approved):
code review: lgtm

PR #54, opened 2019-08-22
[11.0][FIX] .travis.yml: Makepot

astirpe:


astirpe:
@pedrobaeza can we fast track this?

PR #55, opened 2019-08-22
[12.0][FIX] .travis.yml: Makepot

astirpe:


astirpe:
@pedrobaeza can we fast track also this one please? Thank you!

PR #57, opened 2019-09-09
12.0 mig l10n de skr03 mis reports

OSevangelist:
MIgration of Module


OSevangelist:
@fkantelberg can you please review what goes wrong with my PR and the check and why it exactly fails. Runbot says something with the mis_builder but that module already exists in v12

fkantelberg:
@OSevangelist 
The misbuilder module moved to OCA/mis-builder which means you have to update the file oca_dependencies.txt in the root of the repository. The CI is loading it to check which further OCA repos he has to include. You should also keep care of the comment in the manifest.

OSevangelist:
@fkantelberg fixed one issue and directly got the next. It has to do with the country states but i have no clue how this could potentially go wrong. Any idea

fkantelberg:
@OSevangelist This one is weird and I don't see any hint in the code why this would even happen except that the previous installation went a bit wrong. I would suggest a debugging in the container. Check out [this](https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#investigating-travis-test-failures) It looks like some remnants and you can recreate the database.

l10n_de_skr03_mis_reports/__manifest__.py
```diff
@@ -0,0 +1,25 @@
+# -*- coding: utf-8 -*-
```
daramousk: Please remove the encodings from all your *.py files.

daramousk (changes requested):
Hi there, thanks for this,

I think that travis is broken here, can you please make a separate commit over here that changes the .travis file to look like the one here https://github.com/OCA/maintainer-quality-tools/blob/master/sample_files/.travis.yml (follow the instructions there as well.

We need to make sure that travis works.

marylla:
Hi @OSevangelist,
could you solve the problem?

OSevangelist:
@fkantelberg would please be so kind and help me on getting rid of he Runbot and Travis errors here once again. It seems that i am not the only person on this repository having these issues. In order to keep up a bit and improve the status of the German localisation it would be nice if we resolved the issues one by one

fkantelberg:
@OSevangelist 
Yeah no wonder the country states are defined twice. Once in the module `l10n_de_country_states` and once in the `l10n_de` module. See https://github.com/OCA/OCB/blob/12.0/addons/l10n_de/data/res.country.state.csv but this file is only in version 12. Which makes the module l10n_de_country_states deprecated BUT the country states aren't ported to v13+ yet in base odoo.

marylla:
@fkantelberg Can you please re-check this PR?

fkantelberg:
@marylla Nothing has changed in the 12.0 branch which means he would still fail. Technically the errors have nothing to do with this MR and #98 wanted to take care of it. I think we can follow the advice there and make the `l10n_de_country_states` uninstallable for 12.0

PR #64, opened 2020-02-17
updated module l10n_de_skr04 for Odoo v12

DavidNitro:


tv-openbig:
@jans23 I have just checked the modules upgrade by hand on the created runbot.
To my opinion it works fine, however i have now idea why runbot fails and the travis checker didn't start. By hand it was possible to install, the templates was created and after some account moves also the values was present in the report.

@pedrobaeza Is there any possibility to bypass the automatic checks somehow. I guess it is really difficult to automatically test such localisation apps as we obviously would have to configure a new german company at first. 

Bildschirmfoto 2020-02-18 um 10 07 37

PR #66, opened 2020-02-26
13.0 mig l10n de steuernummer

AMhadhbi:
* Migration of module l10n_de_steuernummer


tbaden (approved):
did a code review: lgtm

but some may not want to change from ' to " even if if unifies the code, you can't look for specific snippets from older version, for some patch commits for example.

hkapatel-initos (changes requested):
We need to follow the guideline to commit. You can have a look at the below link
 https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-13.0
Please keep Migration and Pre-commit separate for 13.0.

hkapatel-initos:
Hi, @AMhadhbi Can you Please keep Migration and Pre-commit separate for 13.0.

hkapatel-initos:
Hello, @tv-openbig Can you please do a functional test for this module?

hkapatel-initos:
Hello, @tv-openbig Can you please do a functional test for this module?

tv-openbig:
@hkapatel-initos Is anybody able to explain the sense of this module ? The field "vat" in "res.partner" is quite clear. We don't need a validator for a totally different number, which should be used in a wrong context of the "vat" field. If anybody needs also the "Steuernummer" then they please should create a module which extends the model "res.partner" by a new field, i.e. "tax_no". For what purpose ever they need this new field, i can imagine to inject it on the invoices in the header or footer   they also would have to change the reports. However it cannot be that a field which is already handled in odoo standard by a validator should be misused by checking the validity of a totally different number. 

I will not merge that module, until anybody is there to explain the sense of it. Unfortunately we have merged this module in the history. This was already a mistake, but it was initiated by a well known foreign developer who obviously didn't knew the difference between vat_id and tax_id in germany.

hkapatel-initos:
Hi, @AMhadhbi Can you please check the above comment and it would be if you take part in this conversation.

hkapatel-initos:
Hi, @AMhadhbi Can you please check the above comment and it would be if you take part in this conversation.

dsolanki-initos:
@AMhadhbi it would be nice if you check these comments.

dsolanki-initos:
@AMhadhbi it would be nice if you check these comments.

tv-openbig:
> @AMhadhbi it would be nice if you check these comments.

@dsolanki-initos Sorry, let's close that PR.

PR #69, opened 2020-05-08
[IMP] show partner details on zm-line validation

Guenzn:
When you post a statement, the zm lines will be created and validated. If there the country codes of the lines are invalid you get a warning, that the country code is not allowed. With this PR, you will also see the name, reference and vat of the partner at the dialog.


tv-openbig (approved):
Thanks for this very usefull extension.

PR #70, opened 2020-06-04
[11.0] added zm export functionality

Guenzn:
Go to the zm page and just click on the download button to get a valid csv file for elster online


tv-openbig:
I have seen on a live instance that the download is working.

tv-openbig (approved):
The download is working.

PR #71, opened 2020-06-08
[FIX] Invalid license key

dreispt:


rvalyi (approved):

l10n_de_country_states/__openerp__.py
```diff
@@ -38,5 +38,5 @@
         'data/res_country_states.xml'
     ],
     'installable': True,
-    'license': 'GPL-3 or any other later version',
+    'license': 'GPL-3',
```
pedrobaeza: https://github.com/odoo/odoo/blob/9e8f70e4849b0eeaca8b5cf51372ecfa23dc561b/openerp/addons/base/module/module.py#L304

```suggestion
    'license': 'GPL-3 or any later version',
```

PR #94, opened 2021-10-05
[12.0][mig]l10n_de_skr03_mis_reports

hkapatel-initos:


hkapatel-initos:
Hi, @tv-openbig the country states are defined two times in this module. Once in the module l10n_de_country_states and once in the l10n_de module. this is the link  https://github.com/OCA/OCB/blob/12.0/addons/l10n_de/data/res.country.state.csv and this is in the 12.

PR #98, opened 2021-11-02
[FIX] state l10n_de_country_states

fshah-initos:


hkapatel-initos:
Hi, @tv-openbig can we merge this?

tv-openbig:
Hello @hkapatel-initos,

The issue is the following. Odoo have introduced in v12 l10n_de a data file 'res.country.state.csv':
Bildschirmfoto 2021-11-12 um 18 23 20

In v13 l10n_de project this data file isn't available anymore:
Bildschirmfoto 2021-11-12 um 18 26 01

But we introduced in the OCA l10n-germany the module already for v11:


Bildschirmfoto 2021-11-12 um 18 27 09


Finally the migration from v11 to v12 leads for sure to conflicts with the Odoo standard if someone have installed l10n_de module which are most german odoo users ....

So if we would merge your module, i assume there will be conflicts, especially if someone will install l10n_de later on. I assume there is a constraint on 'code' field.

It is difficult for me to decide about that merge request.
We need further help to decide what to do: @fkantelberg

fshah-initos:
@tv-openbig Can you check this comment in the previous PR, 
https://github.com/OCA/l10n-germany/pull/57#issuecomment-664144487
it would be nice if we can merge this ASAP.

fkantelberg:
I don't know what the PR should even fix. Removing the first of the entry will just move the error to the next one. => Not a fix at all. In the end this module shouldn't be installable in v10-v12 because Odoo also pushed the same codes. Maybe a port within l10n_de is the best thing and most consistent.

fshah-initos:
I think the "l10n_de_country_states" module should be installable "False" in v10-v12 and it should be installable "true" in v13-v14.

PR #99, opened 2021-11-15
Resolve the error "unescaped double quotes"

mrgoetz:
In the lines 108 and 1035 in de.po of module l10n-de_tax_statement we find "unescaped double quotes". To avoid this error write &quot; instead of " inside the quotes.


mrgoetz:
Wrong solution by a typo.

PR #117, opened 2023-03-06
[14.0][ADD] datev_export_xml

fkantelberg:
This module adds an export of invoices towards the DATEV XML format usable by tax consultants.


chandni299:
Hello @fkantelberg Locally I tested the module/PR and ran the Unittests, It shows some failures like
1.
```
File ".../oca-l10n-germany/datev_export_xml/tests/test_datev_export.py", line 316, in _run_test_document
    doc(".//property[@key='InvoiceType']").attrib["value"],
AttributeError: 'NoneType' object has no attribute 'attrib'
```
2.
```
File ".../oca-l10n-germany/datev_export_xml/tests/test_datev_export.py", line 313, in _run_test_document
    self.assertEqual(doc(".//extension[@name]").attrib["name"], inv_number + ".pdf")
AttributeError: 'NoneType' object has no attribute 'attrib'
```
3.
```
File ".../oca-l10n-germany/datev_export_xml/tests/test_datev_export.py", line 310, in _run_test_document
    doc(".//extension[@datafile]").attrib["datafile"], inv_number + ".xml"
AttributeError: 'NoneType' object has no attribute 'attrib'
```
But all tests are running without any errors [here](https://app.travis-ci.com/github/OCA/l10n-germany/jobs/597746183#L1620)
So I am not sure if it is relevant here or not.

fkantelberg:
@chandni299 Are you running the tests with the demo data of the module?

chandni299:
yes. Unit tests from [https://github.com/OCA/l10n-germany/blob/79a49d549be5b246925b0e6bad4dc461e632752d/datev_export_xml/tests/test_datev_export.py](https://github.com/OCA/l10n-germany/blob/79a49d549be5b246925b0e6bad4dc461e632752d/datev_export_xml/tests/test_datev_export.py) are using demo data if I am not wrong

fkantelberg:
I guess it's a conflict with a local database with mixed demo data and custom data because some of the tests are generating files on all available invoices which will cause unexpected results for the unittests. But I can't fully pinpoint it.

.pre-commit-config.yaml
```diff
@@ -101,7 +101,7 @@ repos:
       - id: pyupgrade
         args: ["--keep-percent-format"]
   - repo: https://github.com/PyCQA/isort
-    rev: 5.5.1
+    rev: 5.12.0
     hooks:
       - id: isort
         name: isort except __init__.py
```
hbrunn: please do this kind of thing in a separate PR in the future, because this we can fasttrack to avoid conflicts and have all other PRs rebase afterwards

datev_export/models/res_company.py
```diff
@@ -0,0 +1,20 @@
...
+from odoo import fields, models
+
+
+class ResCompany(models.Model):
+    _inherit = "res.company"
+
+    datev_consultant_number = fields.Char(
+        string="Consultant Number",
+        size=32,
+        help="Number from 1000 to 9999999",
```
hbrunn: why don't you restrict the field size to the the allowed range?

datev_export/models/res_company.py
```diff
@@ -0,0 +1,20 @@
...
+    )
+
+    datev_client_number = fields.Char(
+        string="Client Number",
+        size=32,
+        help="Number from 0 to 99999",
```
hbrunn, line 19: same here

hbrunn (approved):

datev_export/models/res_company.py
fkantelberg: I limitted the size a bit further. Wouldn't it be better to use an Integer here or atleast a regex constraint? I don't find anything official talking about the bounds of these numbers. What do you think?

.pre-commit-config.yaml
fkantelberg: I rebased.

wildi1:
Hello @tv-openbig,
I just tested it locally with demo data. All tests succeeded but there was one warning:
`odoo_1|2023-03-14 18:50:43,787 1 WARNING devel odoo.addons.datev_export_xml.models.datev_xml_generator: b'<invalid/>'`

One more thing I have noticed: 
The "Set Done" button has no effect in state "pending". Perhaps we could hide it in this state.
image

wildi1 (commented):
Hello @tv-openbig,
I just tested it locally with demo data. All tests succeeded but there was one warning:
`odoo_1|2023-03-14 18:50:43,787 1 WARNING devel odoo.addons.datev_export_xml.models.datev_xml_generator: b'<invalid/>'`

One more thing I have noticed: 
The "Set Done" button has no effect in state "pending". Perhaps we could hide it in this state.
image

fkantelberg:
@wildi1 Thanks for the testing. We can ignore the warning because it also tested garbage input and this is one of it.

Good catch with the button the states between the python code and view were off. The button is now visible in the states `running` and `failed`.

datev_export_xml/i18n/de.po.
hbrunn, line 1: @fkantelberg I just discovered the dot after `po`, you probably should `git mv` that

datev_export_xml/i18n/de.po
```diff
@@ -0,0 +1,954 @@
...
+"                                <span class=\"fa fa-lg fa-building-o\" title=\"Values set here are company-specific.\" aria-label=\"Values set here are company-specific.\" groups=\"base.group_multi_company\"/>"
+msgstr "<span class=\"o_form_label\">Beraternummer</span>\n"
+"                                <span class=\"fa fa-lg fa-building-o\" title=\"Diese Werte sind unternehmensspezifisch.\" aria-label=\"Values set here are company-specific.\" groups=\"base.group_multi_company\"/>"
+
+#. module: datev_export_xml
+#: model_terms:ir.ui.view,arch_db:datev_export_xml.res_config_settings_view_form
+msgid ""
+"<span class=\"o_form_label\">Customer Order Reference</span>\n"
+"                                <span class=\"fa fa-lg fa-building-o\" title=\"Values set here are company-specific.\" aria-label=\"Values set here are company-specific.\" groups=\"base.group_multi_company\"/>"
+"<span class=\"o_form_label\">Kunden Bestellreferenz</span>\n"
```
hbrunn: @fkantelberg this line is missing the `msgstr` clause

PR #119, opened 2023-03-20
[FIX] l10n_de_tax_statement: Escape double quotes in PO file

hbrunn:


hbrunn:
@OCA/local-germany-maintainers could you fasttrack this as it blocks merging v13 PRs?

PR #122, opened 2023-05-05
[FIX] l10n_de_tax_statement - l10n_de_tax_invoice_basis readonly=False in res.config.settings

ghost:
Right now, if you want to change the config "DE Tax Invoice Basis", you can not do it in the settings wizard since the field is readonly.

### DIFF
--- a/l10n_de_tax_statement/models/res_config_settings.py
+++ b/l10n_de_tax_statement/models/res_config_settings.py
@@ -11,4 +11,5 @@ class ResConfigSettings(models.TransientModel):
     l10n_de_tax_invoice_basis = fields.Boolean(
         string="DE Tax Invoice Basis",
         related="company_id.l10n_de_tax_invoice_basis",
+        readonly=False,
     )


hbrunn (approved):

tv-openbig:
@jdidderen-noviat @hbrunn Sorry this field have to be readonly and it should not be changed as otherwise the module l10n_de_tax_statement won't work properly anymore.

hbrunn:
but how are users supposed to configure this in the first place? would a constraint help forbidding to change this when there are already moves (posted) in the company?

tv-openbig:
@jdidderen-noviat @hbrunn I am only able to provide a bit background. The module was an adoption of l10n_nl_tax_statement. In netherlands they need this option explained here in the readme: https://github.com/OCA/l10n-netherlands/tree/14.0/l10n_nl_tax_statement#readme

It is exactly related to this configuration: Go to Invoicing -> Configuration -> Settings, enable/disable NL Tax Invoice Basis (Factuurstelsel) and Apply

We have decided in the past that we just need the equivalent for "Factuurstelsel" in germany. Therefore the other option "deactivated boolean" is irrelevant for germany (in netherlands this would be "Kasstelsel"). 

I assume that we removed in the past all the source code (different dates) for the "Kasstelsel" scenario and just kept the source code for the "Factuurstelsel" option. As we needed to keep the source code for the "Factuurstelsel" option, we also needed to keep that setting.

To my opinion this setting is not required anymore, if we would adopt the overall source code that there is nothing left wihich relates to this standard  "Factuurstelsel" option. 

I would have one remaining open question which i would not be able to answer at that time. There is a Odoo standard cash basis setting, which means: "creation of tax journal lines at the moment that a payment is received". In germany this is called "Istversteuerung". We would need that theoretically too if a company decides for that to create a proper tax declaration (based on the payment dates of taxes with this option activated) .

In the netherlands readme there is this part: This setting is not compatible with the Odoo Cash basis (Kasstelsel) setting which leads to the creation of tax journal lines at the moment that a payment is received, so when using Cash basis you need to disable the Invoice basis setting to create a valid Dutch tax declaration using this module.

Therefore we would need to deactivate the option if we want to do the "Istversteuerung" in germany, but to my opinion that is not true (so maybe a misunderstanding of this part from the netherlands Readme).

 

, so when using Cash basis you need to disable the Invoice basis setting to create a valid Dutch tax declaration using this module.

hbrunn:
@jdidderen-noviat can you describe in which scenario you need to edit this? Judging from

https://github.com/search?q=repo%3AOCA%2Fl10n-germany%20l10n_de_tax_invoice_basis&type=code

it indeed looks like this flag is not used anywhere

ghost:
@hbrunn @tv-openbig  Sorry for the late answer but our customer wants the tax declaration based on the accounting date and not the invoicing date and they do not want to add manually the invoices outside of the period with the tab "Past Undeclared Invoices". 

In this case, they missed some invoices from December 2022 so they encoded it in Odoo with the invoice date of the vendor invoice (so December 2022) and with an accounting date in May 2023. Then when creating the tax declaration, they are not included. 

And when I looked in the code of this module, I found the flag in question in the tax declaration so I wanted to edit it in the configuration wizard but was not able.
https://github.com/OCA/l10n-germany/blob/14.0/l10n_de_tax_statement/models/l10n_de_tax_statement.py#L396

tv-openbig:
@jdidderen-noviat @hbrunn Sorry for my late reply on this topic. As we adopted it from l10n_nl_tax_statement we have removed or at least uncommented all source code related to this setting "Kasstelsel". Maybe this was a bad decision as i understand the sense, especially if the filed "date" from acccount.move is visible on the customer invoice. In fact Odoo EE calculates the tax declarations also more in the way of the "Kasstelsel" option. 

For supplier invoices i am currently not 100% sure if it is calculated from the account.move date or from the invoice date of a supplier bill. At least for the supplier bills the account.move date would be the best choice.

I would be glad to offer this option in this module however we would need to refactor the source code again in order to integrate the functions / methods in the context of "Kasstelsel".

Another scenario where it would be required to have it are the "cutoff" modules from OCA, where also the taxes could be moved into another "booking periode".

PR #139, opened 2023-12-14
[MIG] l10n_de_skr04_mis_reports: Migration to 15.0

ozoromo:
Migration of l10n_de_skr04_mis_reports to version 15.0
(Recreated because of branch change)


ozoromo:
It seems pre-commit/flake8 will always fail since it crashes at runtime.

PR #143, opened 2024-01-19
[15.0][MIG] l10n_de _tax_statement

MrGigSolutions:
Migration of l10n_de_tax_statement to 15.0

Changes:
- Adjusted manifest
- Removed migrations
- Refactored unlink() to @api.ondelete
- Removed unused methods and imports from tests
- Tests now use setUpClass instead of setUp
- Chatters now don't have specified widgets anymore

PR #145, opened 2024-01-23
[15.0][MIG] l10n_de_tax_statement_zm

MrGigSolutions:
Migration of l10n_de_tax_statement_zm to Odoo 15.

Small changes needed only:
- Manifest version
- Changed the tests to use setUpClass, instead of setUp.

NOTE: As this module relies on l10n_de_tax_statement, the merge is based on PR https://github.com/OCA/l10n-germany/pull/143.


MrGigSolutions:
The failing pre-commit seems to be an issue with autoflake, perhaps it needs a version increase.

PR #153, opened 2024-03-19
[MIG][16.0] Migration of datev_export_xml

jsierp:


hbrunn (approved):

hbrunn:
don't worry about runboat, that gets stuck sometimes, haven't yet understood the pattern

datev_export_xml/tests/test_datev_export.py
```diff
@@ -154,7 +150,7 @@ def create_out_invoice(self, customer, start_date, end_date):
                             "credit": 600.0,
                             "debit": 0.0,
                             "account_id": self.account_income.id,
-                            "analytic_account_id": self.analytic_account_it.id,
+                            "analytic_distribution": {self.analytic_account_it.id: 1},
```
fkantelberg: It's only a test but these are in percent which means you have 1% set for all those.

datev_export_xml/views/templates.xml
```diff
@@ -32,7 +32,7 @@
 
         <accounting_info
             t-att-account_no="line.account_id.code or ''"
-            t-att-cost_category_id="line.analytic_account_id.code or ''"
+            t-att-cost_category_id="(line.analytic_line_ids[0].account_id.code or '') if line.analytic_line_ids else ''"
```
fkantelberg: This might be problematic if you actually use an analytic distribution or somebody manually set those (in different order) which would mean that sometimes an analytic account is used and sometimes a different one is used because the lines are ordered mainly by date and id.

Maybe a new `doc.datev_analytic_account(line)` function here? To allow an easier hook and custom selection of the "correct" analytic account?

Either way you could shorten it with `line.analytic_line_ids[:1].account_id.code or ''`

datev_export_xml/views/templates.xml
```diff
@@ -85,15 +85,17 @@
...
+                <t t-if="doc.move_type in ['out_invoice', 'out_refund']">
+                    <t t-set="partner" t-value="doc.partner_id" />
+                    <t
+                        t-set="account"
+                        t-value="partner.property_account_receivable_id"
+                    />
+                </t>
+                <t t-else="">
+                    <t t-set="partner" t-value="doc.company_id.partner_id" />
+                </t>
```
fkantelberg: It's an actual change now. Previously account was always set to the receivable of partner. Now it's not in the `t-else` branch. Why do you change the tested logic here when the previous code worked?

Do you have an actual problematic case? Do you find it hard to read with the `t-set t-value t-if` in one tag?

See also https://github.com/OCA/l10n-germany/pull/126#discussion_r1508840646

fkantelberg (changes requested):

fkantelberg (approved):

tv-openbig:
@jsierp @fkantelberg

I am wondering why the smart button "Invoices" (showing the included invoices / bills) is not anymore available ?
Furthermore some fields in the general settings section of the form view datev.export.xml.form are missing.
v14:
Bildschirmfoto 2024-03-27 um 10 49 49
v16:
Bildschirmfoto 2024-03-27 um 10 45 03

jsierp:
@tv-openbig I have fixed the visibility of smart buttons. The other fields are visible only if they are set, and they should be set in the settings. As I understand they are just for some overview and I wouldn't change the behavior now during the migration.

tv-openbig:
@jsierp fields are there and smart button is also there
Bildschirmfoto 2024-05-09 um 13 50 26

PR #154, opened 2024-03-25
[14.0][BACKPORT] datev_export_xml: Backport account no always on invoice_party

fkantelberg:
I got notified that this change was missing in the backports last week. It ensures that the account no is always visible for the `invoice_party`.


fkantelberg:
FYI @tv-openbig I rechecked the entire diff between 14.0 and 15.0 again and there is no other technical difference beside migration stuff.

PR #155, opened 2024-04-05
[IMP][FIX][14.0] datev_export_xml: Validation and small issues

fkantelberg:
**Fixes:**
- When you create an invoice line without product but with a longer description the `description_short` of `invoice_item_list` can get longer than 40 characters
- When you have a discount of 100% or something like free shipping the `net_product_price` of `invoice_item_list` would be 0 but DATEV doesn't allow 0 for those (mysteries of DATEV but I guess because no movement happened between the accounts)

**Improvement:**
It's inconvenient if there are errors in the invoices when rendering the ZIP archive because it stops at the first error. This is why I added a validation button and view with problematic invoices to jump into them and fix them one by one before trying to generate the ZIP.

PR #156, opened 2024-04-08
[IMP][15.0] datev_export_xml: Validation and small issues

fkantelberg:
Forward port of #155

### DIFF
--- a/datev_export_xml/models/datev_xml_generator.py
+++ b/datev_export_xml/models/datev_xml_generator.py
@@ -23,16 +23,23 @@ class DatevXmlGenerator(models.AbstractModel):
     _description = "DATEV XML Generator"
 
     @api.model
-    def check_xml_file(self, doc_name, root, xsd=None):
+    def check_xml_file(self, doc_name, root, xsd=None, invoice=None):
         if not xsd:
             xsd = "Document_v050.xsd"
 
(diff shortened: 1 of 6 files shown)


fkantelberg:
@tv-openbig This is the forward port of #155

datev_export_xml/models/datev_xml_generator.py
```diff
@@ -23,16 +23,23 @@ class DatevXmlGenerator(models.AbstractModel):
...
+            if invoice:
+                invoice.datev_validation = False
+
             schema = etree.XMLSchema(etree.parse(schema))
             schema.assertValid(root)
         except (etree.DocumentInvalid, etree.XMLSyntaxError) as e:
             _logger.warning(etree.tostring(root))
+
+            if invoice:
+                invoice.datev_validation = str(e)
```
pedrobaeza, line 41: This data is not going to be written as the latter raise rollbacks the transaction.

fkantelberg: This is correct for the normal export but not for the validation feature (`action_validate`)

fkantelberg:
FYI our customer found an issue with it. When you have multiple taxes on an invoice it crashes in the qweb and no usable exception is given. Multiple taxes are checked in `_check_invoices` and I added it to each single invoice check.

PR #158, opened 2024-05-13
[14.0][IMP] datev_export_xml: Improve handling and download of bigger files

fkantelberg:
Currently when big files are handled the server might respond with 500 MemoryError when opening the list or form view of the datev xml exports. The worker processes are hitting the RAM limits and will be killed.

PR #159, opened 2024-05-14
[16.0][IMP] datev_export_xml: Improve handling and download of bigger files

fkantelberg:
Currently when big files are handled the server might respond with 500 MemoryError when opening the list or form view of the datev xml exports. The worker processes are hitting the RAM limits and will be killed.

#158


fkantelberg:
This also cherry-picks the missing validation feature.

PR #170, opened 2025-01-31
[17.0][MIG] datev_import_csv_dtvf: Migration to 17.0.

pniederlag:
yet untested


tv-openbig:
@pniederlag I have checked the functionality of the module as you can see.

Wizard:

Bildschirmfoto 2025-02-12 um 17 30 10



Journal entry:
Bildschirmfoto 2025-02-12 um 17 38 59

As in prior versions:
We do not import taxes. Probably these lines needs to be adopted in Odoo by entering the net value and the tax in the lines booked on income accounts.

tv-openbig (commented):
I have tested the functionality on Odoo 17 (EE). See comments.

PR #173, opened 2025-02-18
[17.0][MIG] datev_export: Migration to 17.0

victoralmau:
Migration to 17.0

@Tecnativa TT54914

PR #174, opened 2025-02-19
[17.0][MIG] datev_export_xml: Migration to 17.0

victoralmau:
Migration to 17.0

Locked by:
- [x] `datev_export`: https://github.com/OCA/l10n-germany/pull/173

@Tecnativa TT54915


david-fleischmann:
@victoralmau 
image
Something is not correct yet in that version. I'm unable to see errors in the xml export to fix them. In v14 there was a Validate button where I had a clear overview of the erros in the accoun.move records:
image

david-fleischmann:
@victoralmau @fkantelberg 
Just found that the display of the exception_info seems not to work as in V16 (runbot):
image

In V17 the rendering of the error message seems to fail:
image

victoralmau:
> @victoralmau image Something is not correct yet in that version. I'm unable to see errors in the xml export to fix them. In v14 there was a Validate button where I had a clear overview of the erros in the accoun.move records: image

You are right, actually this was added in v15 but not in v16, therefore, I have added it first in v16 https://github.com/OCA/l10n-germany/pull/179 and when it is merged, I will add it here.

IMO the comment https://github.com/OCA/l10n-germany/pull/174#issuecomment-2781614407 is related to the other change.

pedrobaeza:
/ocabot migration datev_export_xml

Please remove the temporary commit.

datev_export_xml/tests/test_datev_export.py
```diff
@@ -0,0 +1,868 @@
...
+        )
+
+        cls.parent_customer = cls.env.ref("datev_export_xml.customer_parent")
+        cls.child_customer = cls.env.ref("datev_export_xml.customer_child")
+
+        # Ensure the initial state
+        cls.refund_date = cls.today - timedelta(days=55)
+        cls.start_date = cls.today - timedelta(days=34)
+        cls.end_date = cls.today - timedelta(days=32)
+        cls.InvoiceObj.with_context(force_delete=True).search([]).unlink()
```
etobella, line 110: This line is problematic if we have other modules installed. 

For example, if you install sale and try to pass the test, it will fail (no other modules installed). The problem comes from the fact that odoo auto installs account_audit_trail by default on `l10n_de`.

The way to solve it is to add 
``` python
        if "check_account_audit_trail" in cls.env["res.company"]._fields:
            cls.env["res.company"].search([]).write({"check_account_audit_trail": False})
```
Can you add it?

pedrobaeza (approved):

pedrobaeza:
@etobella @david-fleischmann is this OK for you then?

etobella (approved):
LGTM

PR #175, opened 2025-02-19
[17.0][MIG] l10n_de_skr04_mis_reports + l10n_de_skr03_mis_reports > l10n_de_mis_reports: Migration to 17.0

victoralmau:
Migration to 17.0

@Tecnativa TT54925


pedrobaeza:
Does this report serves for both skr04 and skr03?

victoralmau:
Checking the `l10n_de_skr04_mis_reports` module does not add anything specific related to `l10n_de_skr04`, I guess the definition of `mis.report` data will do but I am not an expert in German accounting, so someone will have to confirm this. 
In any case, I think the right approach is this and if new mis reports need to be added they should be added here.

pedrobaeza:
@tv-openbig @OSevangelist what do you think about this?

tv-openbig:
@pedrobaeza @victoralmau Hello,

please compare trhis both files in deepth:
https://github.com/OCA/l10n-germany/blob/14.0/l10n_de_skr03_mis_reports/data/mis_report_pl.xml
https://github.com/OCA/l10n-germany/blob/14.0/l10n_de_skr04_mis_reports/data/mis_report_pl.xml

I guess we still require this both modules especially to the balp formulas, in which we relate to account 
code field.

victoralmau:
Note that I am not saying that the module(s) are not needed, just that now in odoo core there is no `l10n_de_skr04` module, only `l10n_de`, and for that reason it has been renamed to `l10n_de_mis_reports`. Then when migrating `l10n_de_skr03_mis_reports` the data will be added to this module (`l10n_de_mis_reports`).

pedrobaeza:
Yeah, I agree on unifying both modules in one, but let's do that in one shot, Víctor, although we don't use SKR03, for not letting the job half done. Thus, put the history of both modules, and the commit moving them to the common one. 

There's still pending the correction for allocating the incomes/expenses in the balance sheet, that will be handled from 13.0 up to this version.

victoralmau:
> Yeah, I agree on unifying both modules in one, but let's do that in one shot, Víctor, although we don't use SKR03, for not letting the job half done. Thus, put the history of both modules, and the commit moving them to the common one.
> 
> There's still pending the correction for allocating the incomes/expenses in the balance sheet, that will be handled from 13.0 up to this version.

Changes done.
One question, the renaming of all the xmlids where exactly should it be done? I ask because normally I would add it here in migrations/17.0.1.0.0/pre-migration.py, but as there is https://github.com/OCA/OpenUpgrade/pull/4748 maybe it needs to be added in OpenUpgrade.

tv-openbig:
> Note that I am not saying that the module(s) are not needed, just that now in odoo core there is no `l10n_de_skr04` module, only `l10n_de`, and for that reason it has been renamed to `l10n_de_mis_reports`. Then when migrating `l10n_de_skr03_mis_reports` the data will be added to this module (`l10n_de_mis_reports`).

@victoralmau Ok i understand and that makes sense.

tv-openbig:
@victoralmau I stumbled over a issue.

May you change in .../data/mis_report_pl_skr03.xml the following "balp" expressions in the lines:
309
355

remove the ",23%" from the balp expression.

It is conflicting with the line 212. 
That was leading to appearance of account move lines on the account "2315" in three report lines.

Thanks in advance.

Everything else looks good.

tv-openbig:
> There's still pending the correction for allocating the incomes/expenses in the balance sheet, that will be handled from 13.0 up to this version.

@victoralmau @pedrobaeza Yes that is right. It was never implemented by code logic. It is required to create a booking for the unallocated profit/loss account like described here: https://www.odoo.com/de_DE/forum/hilfe-1/how-to-clear-undistributed-profit-loss-account-999999-in-odoo-152229

That is far from perfect.

victoralmau:
> May you change in .../data/mis_report_pl_skr03.xml the following "balp" expressions in the lines: 309 355
> 
> remove the ",23%" from the balp expression.
> 
> It is conflicting with the line 212. That was leading to appearance of account move lines on the account "2315" in three report lines.
> 
Changes done.

victoralmau:
Oops, I think the command is incorrect, can you try again?

PR #179, opened 2025-04-22
[16.0][IMP] datev_export_xml: Validation and small issues

victoralmau:
FWP from 15.0: https://github.com/OCA/l10n-germany/pull/156

Validation and small issues

Related to https://github.com/OCA/l10n-germany/pull/174#issuecomment-2781614407

@Tecnativa

### DIFF
--- a/datev_export_xml/tests/test_datev_export.py
+++ b/datev_export_xml/tests/test_datev_export.py
@@ -20,9 +20,40 @@ class TestDatevExport(TransactionCase):
     @classmethod
     def setUpClass(cls):
         super().setUpClass()
+        cls.company = cls.env.company
         cls.JournalObj = cls.env["account.journal"]
-        cls.sale_journal = cls.JournalObj.search([("type", "=", "sale")])[0]
-        cls.purchase_journal = cls.JournalObj.search([("type", "=", "purchase")])[0]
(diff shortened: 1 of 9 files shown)


datev_export_xml/tests/test_datev_export.py
```diff
@@ -20,9 +20,40 @@ class TestDatevExport(TransactionCase):
     @classmethod
     def setUpClass(cls):
         super().setUpClass()
+        cls.company = cls.env.company
```
pedrobaeza, line 23: The solution for this should be the same as the other: post-install + make sure the CoA is loaded.

victoralmau: Indeed that should be the solution, but it is not. I tried to apply that solution but it involves a lot of changes in the tests and demo data only for this 16.0 version because there are a lot of demo data defined (products, accounting accounts, etc) that make the tests work correctly. I agree that this approach is not the right one, but personally it is not worth all the hours of effort to change it to get the same result (I repeat, this change would only apply to v16 tests).

pedrobaeza: OK then.

pedrobaeza (approved):

etobella (approved):
Technical review

PR #185, opened 2025-05-13
[14.0][UPD] l10n_de_tax_statement_zm: New format

fkantelberg:
As far as I know ELSTER changed the content of the file to drop the extra column of the country code while including it in the VAT column.

### DIFF
--- a/l10n_de_tax_statement_zm/models/l10n_de_tax_statement.py
+++ b/l10n_de_tax_statement_zm/models/l10n_de_tax_statement.py
@@ -203,7 +203,7 @@ class VatStatement(models.Model):
             ",".join(str(i) for i in line)
             for line in self._generate_zm_download_lines()
         )
-        zm_download_base64 = base64.b64encode(zm_download.encode("iso-8859-15"))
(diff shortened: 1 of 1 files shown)


fkantelberg:
@tv-openbig A customer mentioned the change and as far as I see he's refering to the version 2.0 change of the file. As far as the version 3.0 goes there shoudn't be any differences in the content because there are no umlauts/special characters which means that the bytes are identical in both encodings.

Do you agree to the changes?

https://www.elster.de/eportal/helpGlobal?themaGlobal=zmdo_import_eop

tv-openbig:
@fkantelberg Yes the change is required. It is the same as in Odoo EE. 
They are also working on that change (today).

tv-openbig:
@fkantelberg Ich habe die Änderung erfolgreich getestet. Vielen Dank für das Update.

Bildschirmfoto 2025-06-06 um 09 47 19

[Germany_ 2025-06-01 2025-06-30.csv](https://github.com/user-attachments/files/20623957/Germany_.2025-06-01.2025-06-30.csv)

PR #188, opened 2025-05-26
[18.0][MIG] datev_import_csv_dtvf: Migration to 18.0

pniederlag:
upgrade datev_import_csv_dtvf for version 18


tv-openbig:
@pniederlag @hbrunn I have tested the import. Currently we have no onchange imppemented if there is an account with a tax as a default for the account.Please compare the last line. Currently i advise the customers to remove the default tax from the account and to create a manual correction for the required tax in "Miscelleanous" journal. The calculation for the onchange should be "tax included".
Bildschirmfoto 2025-06-06 um 09 59 40

PR #190, opened 2025-07-02
[18.0][MIG] l10n_de_tax_statement: Migration to 18.0

ByteMeAsap:


setup/l10n_de_tax_statement/odoo/addons/l10n_de_tax_statement
CRogos, line 1: Shouldn't the hole setup folder be removed with version 17.0?

l10n_de_tax_statement/views/l10n_de_tax_statement_view.xml
```diff
@@ -122,20 +112,20 @@
...
-                                    <field name="is_readonly" invisible="1" />
+                                    <field name="is_group" column_invisible="1" />
+                                    <field name="is_total" column_invisible="1" />
+                                    <field name="is_readonly" column_invisible="1" />
                                     <field name="code" readonly="1" />
                                     <field name="name" readonly="1" />
-                                    <field name="format_base" invisible="1" />
-                                    <field name="format_tax" invisible="1" />
+                                    <field name="format_base" column_invisible="1" />
+                                    <field name="format_tax" column_invisible="1" />
```
CRogos: This should also be fixed in v17.0. Shouldn't we create a patch for this first and create a new PR based on the fixed version?

CRogos (changes requested):
I am new to LazyTranslate. When should this be used and is this the case here?

setup/l10n_de_tax_statement/odoo/addons/l10n_de_tax_statement
ByteMeAsap: Removed it, thanks !!

ByteMeAsap:
> I am new to LazyTranslate. When should this be used and is this the case here?

Odoo 18 gives a warning "odoo.tools.translate: no translation language detected, skipping translation" if we use the default translation option. Using a lazy translation function allows the translation to take place when the string is rendered to the user, as opposed to translated when the framework interprets the line of code thereby resolving the warning.

CRogos:
@ByteMeAsap could you review my fix in v17 and if merged recreate your PR following the migration guide? 

https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-18.0

e.g. some of the commits should be squashed, and i also would expect a pre-commit cleanup commit before the migration?
image

ByteMeAsap:
> i also would expect a pre-commit cleanup commit before the migration?

There was nothing to commit with the initial pre-commit

ByteMeAsap:
> @ByteMeAsap could you review my fix in v17 and if merged recreate your PR following the migration guide?
 Opened the new PR - https://github.com/OCA/l10n-germany/pull/201

Issue #195, opened 2025-07-24
[14.0] wrong header in Zusammenfassende Meldung Report


tv-openbig:
## Module
l10n_de_tax_statement_zm

## Describe the bug
The latest update of the export file still contains an error.
The header contains 4 columns and the data lines have 3 lines.

## To Reproduce
- Create a invoice from a german company to a EU customer (B2B9
- Create a tax report for the month of the invoices booking date.
- Create the .csv report

It should look like this:

<img width="2049" height="1116" alt="Image" src="https://github.com/user-attachments/assets/0747b279-6c05-4084-8a25-0f653e4a5de1" />

**Expected behavior**
The header should be:

Umsatzsteuer-Identifikationsnummer (USt-IdNr.),Summe (Euro),Art der Leistung

**Additional context**
Here is the description of the latest format changes:
https://www.elster.de/eportal/helpGlobal?themaGlobal=zmdo_import_eop

PR #196, opened 2025-07-25
[FIX][14.0] l10n_de_tax_statement_zm: Adjust header of CSV

fkantelberg:
This fixes the header to only use 3 columns as defined in the newer version of the spec and copies the column names from the spec.

#195 

spec: https://www.elster.de/eportal/helpGlobal?themaGlobal=zmdo_import_eop


fkantelberg:
@tv-openbig This should do it.

tv-openbig:
@fkantelberg test was successful
Bildschirmfoto 2025-07-30 um 11 27 05

PR #197, opened 2025-07-29
[17.0] FIX wrong expression at kpi gewinn_des_jahres and improved translations

GuenterSelbertSAMSA-IT:
[FIX] better and correct translations. Replaced wrong expression kpi gewinn_des_jahres (balance sheets), which was selecting accounts by user_type_id (relation account.account.type) which was replaced since odoo v 16 by a simple selection field name account_type

### DIFF
--- a/l10n_de_mis_reports/README.rst
+++ b/l10n_de_mis_reports/README.rst
@@ -20,7 +20,7 @@ github repository.
 Configuration
 =============
 
-To configure this module, you need to go to 
+To configure this module, you need to go to
 Accounting > Reporting > MIS Reports and create report instance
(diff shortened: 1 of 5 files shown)


GuenterSelbertSAMSA-IT:
@tv-openbig fixes balance sheet templates, can you review and merge please

tv-openbig:
@GuenterSelbertSAMSA-IT I have checked your commit and it delivers the correct result on the balance sheet as well. The unallocated earnings are represented in the balance sheet of the current year, which was not the case in prior versions of this module.
Bildschirmfoto 2025-09-05 um 09 19 52

PR #212, opened 2026-01-09
[19.0][MIG] l10n_din5008_move_name: Migration to 19.0

mt-software-de:
Since the field `l10n_din5008_document_title` of `account.move` was removed and moved into the template. 
I removed the models and added the template. 

grafik


mt-software-de:
@etobella @tv-openbig maybe you can review this PR.

l10n_din5008_move_name/tests/test_din5008.py
```diff
@@ -59,5 +60,7 @@ def test_name(self):
             }
         )
         invoice.action_post()
-        # Check the name of the invoice
-        self.assertIn(invoice.name, invoice.l10n_din5008_document_title)
+        html, _ = self.env["ir.actions.report"]._render_qweb_html(
+            "account.report_invoice_with_payments", invoice.ids
+        )
+        self.assertTrue(f"Invoice {invoice.name}", html)
```
etobella: assertTrue? Shouldn't it be assertIn? obviously the string is True...

mt-software-de: Yes of course. Thx

etobella:
The problem is that with the current change, it is no longer this way... it has probably a span inside :thinking:

mt-software-de:
> The problem is that with the current change, it is no longer this way... it has probably a span inside 🤔

Yes. :-)
Done

l10n_din5008_move_name/tests/test_din5008.py
```diff
@@ -59,5 +63,11 @@ def test_name(self):
             }
         )
         invoice.action_post()
-        # Check the name of the invoice
-        self.assertIn(invoice.name, invoice.l10n_din5008_document_title)
+        html, _ = self.env["ir.actions.report"]._render_qweb_html(
+            "account.report_invoice_with_payments", invoice.ids
+        )
+        expected = (
```
etobella: For safety reasons might be good to add a regex this way we are sure that we have no problems with spaces and so on 😄

mt-software-de:
@etobella i would need the addon also now for v18. Whats your recommendation?
Should i create a new unrelated PR. Or should i create PR for v18 and then use this as base for v19, when v18 is merged?

l10n_din5008_move_name/report/din5008_report.xml
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+    <template
+        id="report_invoice_document"
+        inherit_id="l10n_din5008.report_invoice_document"
+        priority="100"
+    >
+        <xpath expr="//t[@t-set='din5008_document_title']/span" position="after">
+            <span t-if="o and o._name == 'account.move'">
```
luisDIXMIT: By moving the logic from Python, you can still keep the original check. https://github.com/OCA/l10n-germany/blob/7145baa7f9d6d2436baab2724eee047673b65d3b/l10n_din5008_move_name/models/account_move.py#L13
```suggestion
            <span t-if="o and o.is_sale_document(True)">
```

l10n_din5008_move_name/report/din5008_report.xml
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+    <template
+        id="report_invoice_document"
+        inherit_id="l10n_din5008.report_invoice_document"
+        priority="100"
+    >
+        <xpath expr="//t[@t-set='din5008_document_title']/span" position="after">
```
luisDIXMIT, line 8: If you prefer direct reference.
```suggestion
        <xpath expr='//span[@t-if="o and o._name == &apos;account.move&apos;"]' position="after">
```

l10n_din5008_move_name/report/din5008_report.xml
etobella: well, it should be o and o._name == 'account.move' and o.is_sale_document(include_receipts=True)

luisDIXMIT: Yes, sorry. That is the best one.

l10n_din5008_move_name/report/din5008_report.xml
mt-software-de: IMO. `din5008_document_title` should/must be in the xpath. 
Its much more readable and explicit. 

If you insist on it. I would change it to. 
`<xpath expr='//t[@t-set='din5008_document_title']//span[@t-if="o and o._name == &apos;account.move&apos;"]' position="after">`

l10n_din5008_move_name/tests/test_din5008.py
```diff
@@ -0,0 +1,71 @@
...
+                "invoice_line_ids": [
+                    (0, 0, {"product_id": self.consulting.id, "quantity": 1})
+                ],
+            }
+        )
+        invoice.action_post()
+        html, _ = self.env["ir.actions.report"]._render_qweb_html(
+            "account.report_invoice_with_payments", invoice.ids
+        )
+        expected = r"Invoice\s*</span>\s*<span>\s*" + re.escape(invoice.name)
```
luisDIXMIT: To fix tests maybe you have to use a regex expresion like 
```suggestion
        expected = r"\w*" +  r"Invoice\s*</span>\s*<span>\s*" + re.escape(invoice.name) + r"\w*"
```

luisDIXMIT (approved):
Code review, LGTM!

etobella (approved):
/ocabot merge nobump

PR #213, opened 2026-01-12
[19.0][ADD] l10n_din5008_stock_picking_type_name: Display the picking type name

mt-software-de:
This addon displays the picking type name in the subject of the DIN 5008 Stock report.


mt-software-de:
@jans23 @NICO-SOLUTIONS @smaddlsoft could you please review this.

NICO-SOLUTIONS:
@mt-software-de 
I wonder if it might make sense to implement this at the level of the account invoice report rather than specifically in l10n_germany / the DIN 5008 report. That way, the functionality wouldn’t be limited to the German/DIN 5008 layout and could potentially be useful for other invoice reports as well.

I already built something along these lines, including partner-specific settings that are only rendered when configured. If that approach is of interest, I’d be happy to share the implementation for review.
What do you think?

mt-software-de:
> @mt-software-de 
> I wonder if it might make sense to implement this at the level of the account invoice report rather than specifically in l10n_germany / the DIN 5008 report. That way, the functionality wouldn’t be limited to the German/DIN 5008 layout and could potentially be useful for other invoice reports as well.
> 
> I already built something along these lines, including partner-specific settings that are only rendered when configured. If that approach is of interest, I’d be happy to share the implementation for review.
> What do you think?

Yes please always share. :-)
But this is for the delivery slip. But still share what you have.

NICO-SOLUTIONS:
@mt-software-de 
Ah, got it — sorry, I was too quick to jump to a conclusion after seeing the context. 😄

In general, I just try to avoid tying features too closely to DIN 5008 if that isn’t necessary. I’ve seen a few approaches where functionality is coupled directly to the DIN 5008 layout, even though a more generic solution could potentially be useful for other reports as well.

But in this case, you’re right — since this is specifically about the delivery slip and the feature needs to follow the DIN 5008 requirements, tying it to the DIN 5008 report makes sense. Actually, that’s also how I’ve implemented it in my code. ;P
image

Anyway, I think we could perhaps containerize things a bit more and, where possible, keep the functionality tied to the stock reports. This is actually what I had already done in my implementation:
image

Looking at the other features that have been implemented for the stock reports, I think there may be a broader architectural question worth discussing. Maybe we should consider some refactoring here. Personally, I’d rather avoid creating a separate module for every single small feature. I think it would be nicer to have more configurable modules that provide a set of related features and can be enabled/configured as needed.

My initial input was focused on invoices, though, so I misunderstood the context here. I’ll share my implementation soon anyway, including the partner-specific settings for the additional stock report information. I just need to bring it into OCA style first. 🙂

NICO-SOLUTIONS:
@mt-software-de 
This is already covered by l10n_din5008_stock:
https://github.com/odoo/odoo/commit/9cd933e9cbb170ccf44f1119d0a18451f1862be1

Looking at the commit, l10n_din5008_stock already sets din5008_document_title using picking_type_id._get_code_report_name().

So it seems like this feature is already covered there.
image

image

mt-software-de:
> @mt-software-de 
> This is already covered by l10n_din5008_stock:
> https://github.com/odoo/odoo/commit/9cd933e9cbb170ccf44f1119d0a18451f1862be1
> 
> Looking at the commit, l10n_din5008_stock already sets din5008_document_title using picking_type_id._get_code_report_name().
> 
> So it seems like this feature is already covered there.
> image
> 
> image
> 

Yes i have seen this as well Yesterday. They fixed

PR #223, opened 2026-04-16
[19.0][MIG] hr_expense_meal_allowance: Migration to 19.0

CRogos:
### Version & Manifest
- Version bump: `18.0.1.1.2` → `19.0.1.0.0`
- Removed `product_product.xml` from manifest (`product_tag_ids` now in base product form in 19.0)
- Added missing `ir_rule.xml` to manifest `data` list
- Removed legacy `migrations/18.0.1.1.0/` directory

### Model Changes (Odoo 19.0 API)
- **Removed `hr_expense_sheet.py`** — `hr.expense.sheet` model no longer exists in 19.0
- **Moved PDF auto-attach logic** from `_do_create_moves` to `action_post` override
- **Uses `nb_attachment`** (replaces `message_attachment_count`)
- **Fixed state reference**: `"reported"` → `"submitted"` in `_compute_currency_id`

### View Changes
- Fixed xpaths: `action_view_sheet` → `action_submit` (sheets removed in 19.0)
- Added `hasclass('oe_inline')` for `total_amount_currency` xpath disambiguation
- Removed `product_product.xml` view (no longer needed)


MohamedOsman7 (commented):
@CRogos 
1. I encountered a display issue with the "Total" field after setting the customer to a company (is_company=True), after setting Travel Begin and Travel End.

Image

2. When selecting a company contact (not individual) as Customer without setting Travel Begin and Travel End, the following error occurs:

Image

Error details below.

Shouldn't a company contact type be set? If so, then the error message shouldn't look like that, right?










Odoo Server Error

RPC_ERROR
Odoo Server Error

Occured on oca-l10n-germany-19-0-pr223-e8d42a5c3409.runboat.odoo-community.org on model hr.expense on 2026-04-22 12:33:45 GMT

Traceback (most recent call last):
  File "/opt/odoo/odoo/http.py", line 2298, in _serve_db
    return service_model.retrying(serve_func, env=self.env)
  File "/opt/odoo/odoo/service/model.py", line 188, in retrying
    result = func()
  File "/opt/odoo/odoo/http.py", line 2353, in _serve_ir_http
    response = self.dispatcher.dispatch(rule.endpoint, args)
  File "/opt/odoo/odoo/http.py", line 2568, in dispatch
    result = self.request.registry['ir.http']._dispatch(endpoint)
  File "/opt/odoo/odoo/addons/base/models/ir_http.py", line 355, in _dispatch
    result = endpoint(**request.params)
  File "/opt/odoo/odoo/http.py", line 808, in route_wrapper
    result = endpoint(self, *args, **params_ok)
  File "/opt/odoo/addons/web/controllers/dataset.py", line 32, in call_kw
    return call_kw(request.env[model], method, args, kwargs)
  File "/opt/odoo/odoo/service/model.py", line 97, in call_kw
    result = method(recs, *args, **kwargs)
  File "/opt/odoo/addons/web/models/models.py", line 2145, in onchange
    record._apply_onchange_methods(field_name, result)
  File "/opt/odoo/odoo/orm/models.py", line 6979, in _apply_onchange_methods
    res = method(self)
  File "/mnt/data/odoo-addons-dir/hr_expense_meal_allowance/models/hr_expense.py", line 150, in _onchange_customer
    ("expire_on", ">=", record.travel_end.date()),
AttributeError: 'bool' object has no attribute 'date'

The above server error caused the following client error:
RPC_ERROR: Odoo Server Error
    RPC_ERROR
        at makeErrorFromResponse (http://oca-l10n-germany-19-0-pr223-e8d42a5c3409.runboat.odoo-community.org/web/assets/b027da7/web.assets_web.min.js:3197:165)
        at XMLHttpRequest.<anonymous> (http://oca-l10n-germany-19-0-pr223-e8d42a5c3409.runboat.odoo-community.org/web/assets/b027da7/web.assets_web.min.js:3204:13)

CRogos:
@MohamedOsman7 I've updated the code and increased the test coverage.

MohamedOsman7 (approved):
functional & code review LGTM

MohamedOsman7:
@pniederlag can you please make a review?

CRogos:
@madmooose could you also check the v19 version?

CRogos:
@NICO-SOLUTIONS if you have time to review.

hr_expense_meal_allowance/data/product_data.xml
```diff
@@ -0,0 +1,18 @@
...
+    <record id="product_tag_meal_allowance" model="product.tag">
+        <field name="name">Expense / Meal Allowance</field>
+    </record>
+    <record id="product_meal_allowance" model="product.template">
+        <field name="name">Meal Allowance</field>
+        <field name="type">service</field>
+        <field name="default_code">VMA</field>
+        <field name="can_be_expensed">True</field>
+        <field name="expense_policy">cost</field>
+        <field name="taxes_id" eval="[]" />
```
NICO-SOLUTIONS:
```suggestion
        <field name="supplier_taxes_id" eval="[]" />
```

NICO-SOLUTIONS: @CRogos 
Shouldn't this use supplier_taxes_id instead of taxes_id for expense products? I believe taxes for expenses are derived from supplier_taxes_id.

NICO-SOLUTIONS (changes requested):
Just a small remark...

hr_expense_meal_allowance/data/product_data.xml
CRogos: You are right, but there are not taxes anyway. I'll just delete it.

hr_expense_meal_allowance/views/hr_expense.xml
```diff
@@ -0,0 +1,78 @@
...
+                        </group>
+                    </page>
+                    <page name="included_meals" string="Included Meals">
+                        <field
+                            nolabel='1'
+                            name="meal_allowance_ids"
+                            readonly="not is_editable"
+                        >
+                            <list editable="bottom" create="0" delete="0">
+                                <field name="date" widget="date" />
```
madmooose, line 39: I suggest to do this readonly. Not sure if we should even do a check on save if this is in lien with the travel date.

NICO-SOLUTIONS (approved):
@CRogos 
@madmooose 

Functional & code review: Looks good to me.

Just one additional thought: it could be worth considering travel stations/stops as part of the meal allowance calculation. That would better represent trips that include multiple locations within the same country, especially when different rates or travel-day calculations come into play.

Image

Image

I've already talked about this with Crogos, but I'd be curious to hear the team's thoughts on it as well.

hr_expense_meal_allowance/report/expense_included_meal.xml
```diff
@@ -0,0 +1,193 @@
...
+                                        </th>
+                                        <th
+                                            name="th_dinner_included"
+                                            class="text-start"
+                                        >
+                                            <span>Dinner Included?</span>
+                                        </th>
+                                        <th
+                                            name="th_expense_for_day"
+                                            class="text-start"
```
madmooose: should be text-end (to be in line with the content)

hr_expense_meal_allowance/report/expense_included_meal.xml
```diff
@@ -0,0 +1,193 @@
...
+                                    <div
+                                        class="m-0"
+                                        t-field="o.travel_begin"
+                                    >2025-01-03</div>
+                                </div>
+                                <div class="col" t-if="o.travel_end" name="travel_end">
+                                    <strong>Travel End</strong>
+                                    <div
+                                        class="m-0"
+                                        t-field="o.travel_end"
```
madmooose: t-options='{"widget": "date"}'

hr_expense_meal_allowance/report/expense_included_meal.xml
```diff
@@ -0,0 +1,193 @@
...
+                                </div>
+                                <div
+                                    class="col"
+                                    t-if="o.travel_begin"
+                                    name="travel_begin"
+                                >
+                                    <strong>Travel Begin</strong>
+                                    <div
+                                        class="m-0"
+                                        t-field="o.travel_begin"
```
madmooose: t-options='{"widget": "date"}'

madmooose (commented):
What about DIN 5008 reports.

This information is missing there

Image

madmooose:
When copying a Meal Expense I get the following error on save
grafik

madmooose:
> @CRogos @madmooose
> 
> Functional & code review: Looks good to me.
> 
> Just one additional thought: it could be worth considering travel stations/stops as part of the meal allowance calculation. That would better represent trips that include multiple locations within the same country, especially when different rates or travel-day calculations come into play.
> Image Image
> 
> I've already talked about this with Crogos, but I'd be curious to hear the team's thoughts on it as well.

I like the idea but I'm not sure if we want to have two customers on one purchase receipt. We could not clearly reinvoice this expenses to customers anymore. Is this a use-case?

NICO-SOLUTIONS:
@madmooose 
That's a valid concern. However, from my POV the primary use case here is internal expense accounting and reimbursement.

One challenge with calculating meal allowances per expense or per travel station is that travel-day calculations can become incorrect, since the allowance is typically determined based on the full travel day and itinerary rather than individual expenses or stops.

We should definitely evaluate the impact on customer re-invoicing, but I would be hesitant to optimize for that use case if it results in less accurate meal allowance calculations.

CRogos:
> What about DIN 5008 reports.

Sorry but I do not support/use the bullshit implementation of DIN 5008. I use css to change the layout to DIN 5008 on the default templates. If someone needs it, he should build a glue module.

> One challenge with calculating meal allowances per expense or per travel station is that travel-day calculations can become incorrect, since the allowance is typically determined based on the full travel day and itinerary rather than individual expenses or stops´

I share this concern, but we should find a way that the correct amount is calculated, but i think we should keep a meal allowance for each customer, to keep the possibility to reinvoice.

The error on duplicate I cannot reproduce.

hr_expense_meal_allowance/report/expense_included_meal.xml
CRogos: This should be datetime not only date.

hr_expense_meal_allowance/views/hr_expense.xml
CRogos: The user must be able to change the values by checking the meals. I think read only is not possible. 
We could add a validation or a constraint?

madmooose: It's possible to add the readonly on the date field. Of course this doesn't catch api interaction but at least accidental user input.

hr_expense_meal_allowance/report/expense_included_meal.xml
madmooose: Alright. Have seen that it's also in a datetime widget when choosing the date.

madmooose:
> > What about DIN 5008 reports.
> 
> Sorry but I do not support/use the bullshit implementation of DIN 5008. I use css to change the layout to DIN 5008 on the default templates. If someone needs it, he should build a glue module.
> 
> > One challenge with calculating meal allowances per expense or per travel station is that travel-day calculations can become incorrect, since the allowance is typically determined based on the full travel day and itinerary rather than individual expenses or stops´
> 
> I share this concern, but we should find a way that the correct amount is calculated, but i think we should keep a meal allowance for each customer, to keep the possibility to reinvoice.
> 
> The error on duplicate I cannot reproduce.

Duplicate the record -> Change the travel date.

madmooose:
> The error on duplicate I cannot reproduce.

Duplicate the record -> Change the travel date.

Full error:

> l10n-germany-19 odoo.sql_db: bad query: b'INSERT INTO "hr_expense_meal_allowance" ("breakfast_included", "create_date", "create_uid", "dinner_included", "hr_expense_id", "lunch_included", "write_date", "write_uid") VALUES (false, \'2026-06-01T12:51:27.927162\'::timestamp, 2, false, 4, false, \'2026-06-01T12:51:27.927162\'::timestamp, 2), (false, \'2026-06-01T12:51:27.927162\'::timestamp, 2, false, 4, false, \'2026-06-01T12:51:27.927162\'::timestamp, 2), (false, \'2026-06-01T12:51:27.927162\'::timestamp, 2, false, 4, false, \'2026-06-01T12:51:27.927162\'::timestamp, 2), (false, \'2026-06-01T12:51:27.927162\'::timestamp, 2, false, 4, false, \'2026-06-01T12:51:27.927162\'::timestamp, 2), (false, \'2026-06-01T12:51:27.927162\'::timestamp, 2, false, 4, false, \'2026-06-01T12:51:27.927162\'::timestamp, 2) RETURNING "id"'
> ERROR: null value in column "date" of relation "hr_expense_meal_allowance" violates not-null constraint
> DETAIL:  Failing row contains (16, 4, 2, 2, null, f, f, f, 2026-06-01 12:51:27.927162, 2026-06-01 12:51:27.927162).

hr_expense_meal_allowance/report/expense_included_meal.xml
madmooose: resolved

NICO-SOLUTIONS:
> > What about DIN 5008 reports.
> 
> Sorry but I do not support/use the bullshit implementation of DIN 5008. I use css to change the layout to DIN 5008 on the default templates. If someone needs it, he should build a glue module.
> 
> > One challenge with calculating meal allowances per expense or per travel station is that travel-day calculations can become incorrect, since the allowance is typically determined based on the full travel day and itinerary rather than individual expenses or stops´
> 
> I share this concern, but we should find a way that the correct amount is calculated, but i think we should keep a meal allowance for each customer, to keep the possibility to reinvoice.
> 
> The error on duplicate I cannot reproduce.

@CRogos

I think it should be sufficient to inherit hr_expense.report_expense and use:
`<xpath expr="//div[hasclass('row', 'o_header')]" position="replace">`

Then replace it with the same code you already have in the div id="informations" from the report_expense_meal_allowance template.

Additionally, I would add:
`<t t-set="din5008_document_information"/>`
to clear the DIN5008 information block.

This should ensure that the information is also reflected in the 5008 layout without creating a dependency on l10n_din5008.

Otherwise I agree — a glue module could make sense, but I think the approach above should work.

If you go for that approach, I think creating a separate template like:
`<template id="report_expense_travel_information">
    <div class="row o_header">
        ...
    </div>
</template>`
    
and then calling it with:
`<t t-call="your_module.report_expense_travel_information"/>`

hr_expense_meal_allowance/report/expense_included_meal.xml
CRogos: I've changed also the text to "Amount"

CRogos:
@NICO-SOLUTIONS The PDF was incorrect in many ways. I've refactored the PDF template. Could you have a look? 

Do you have an idea why the header on the second page is not showing up in din 5008?

image

DIN 5008
image

Non DIN 5008
image

NICO-SOLUTIONS:
> @NICO-SOLUTIONS The PDF was incorrect in many ways. I've refactored the PDF template. Could you have a look?
> 
> Do you have an idea why the header on the second page is not showing up in din 5008?
> 

The h2 element is hidden by CSS (display: none), which explains why the title is not visible in the rendered report.

I noticed that hr_expense_meal_allowance.action_report_hr_expense_meal_allowance is no longer referenced in the action post and print flow due to the removal of hr_expense_sheet.

In addition, the travel information is now being printed on expense reports that are not meal allowances. The check whether `o.meal_allowance_ids` exists before rendering the travel information is not working.

div class="page o_content_pdf" is now created again inside of the main div class="page o_content_pdf" block.

image

mt-software-de:
@mt-software-de

mt-software-de:
@CRogos Thx for the migration.
Are the commits https://github.com/OCA/l10n-germany/pull/223/changes/db753146b9e4bf5a4fcda229e44ec255e2a500f2 and https://github.com/OCA/l10n-germany/pull/223/changes/c232ffe089b1e47cc214177e45c668046f1b6ee4 necessary for the migration? If not i would prefer to create dedicated Pull Requests for it, if they are needed then the should be included in the `[MIG]` commit.

CRogos:
This module is very new and is developed parallel to our migration. The two commits were findings during the testing. 

I think we should integrate https://github.com/OCA/l10n-germany/commit/c232ffe089b1e47cc214177e45c668046f1b6ee4 into the migration. 

For the other Commit I've created an backport for v18 https://github.com/OCA/l10n-germany/pull/254

If you agree with the port, I think we should also leave the commit in this migration. 
I agree this is not 100% perfect, but I currently don't have the time to change and re-test everything.

mt-software-de:
> This module is very new and is developed parallel to our migration. The two commits were findings during the testing.
> 
> I think we should integrate [c232ffe](https://github.com/OCA/l10n-germany/commit/c232ffe089b1e47cc214177e45c668046f1b6ee4) into the migration.
> 
> For the other Commit I've created an backport for v18 #254
> 
> If you agree with the port, I think we should also leave the commit in this migration. I agree this is not 100% perfect, but I currently don't have the time to change and re-test everything.

For me it is fine to keep the commits in, we can do a backport also if they in this PR. It is just not the convention.
But for https://github.com/OCA/l10n-germany/commit/c232ffe089b1e47cc214177e45c668046f1b6ee4 is there really nothing which needs to be backported to v18?

CRogos:
There have been some changes in Odoo with the expense sheets. Therefore the report is not 1:1 portable.

smaddlsoft (approved):
Functionally, LGTM.
Tested:
- creation of expense of type meal allowance
- choosing of client results in correct meal allowance based on the city and country of client
- printing of expense report PDF, on the second page it has a nice overview of the meal allowance and the inclusions and exclusions of meals. This PDF lands in the accounting and ultimately e.g. in Datev export, helpful to document the meal allowance also to e.g. tax advisor.

What I noticed:
- the PDF that is printed and downloaded differs from the PDF that is stored as attachment and ultimately passed on to accounting. In the PDF stored, the first page is blank largely.

Good enough to be merged; I suppose, as the core seems working fine at first glance, maybe the PDF issue can be fixed later, if need be.

hr_expense_meal_allowance/models/hr_expense.py
```diff
@@ -0,0 +1,278 @@
...
+        """Render the meal allowance report as PDF and attach it once per expense."""
+        # sudo is required: core forbids adding attachments to an approved
+        # expense, and this runs post-approval. The existence check must also
+        # run as sudo so it sees the previously generated attachment
+        # regardless of the current user's access to it. It must be a real
+        # search (not nb_attachment, whose compute has no depends and stays
+        # stale for the whole transaction) so a second invocation in the same
+        # transaction cannot create a duplicate.
+        attachment_model = self.env["ir.attachment"].sudo()
+        for expense in self.filtered("is_meal_allowance"):
```
mt-software-de, line 250: Its better to loop only once over the recordset. 
```suggestion
        for expense in self:
            if not expense.is_meal_allowance:
                continune
```

hr_expense_meal_allowance/models/hr_expense.py
```diff
@@ -0,0 +1,278 @@
...
+            )
+            attachment_model.create(
+                {
+                    "name": attachment_name,
+                    "raw": pdf_content,
+                    "res_model": "hr.expense",
+                    "res_id": expense.id,
+                    "mimetype": "application/pdf",
+                }
+            )
```
mt-software-de, line 273: I would not check if a attachment already exists, because when is this executed on approve. 
Also i would stick to the configured names of the report if there are any. Also i would let the attachment be stored the default way if configured. 

```suggestion
        attachment_model = self.env["ir.attachment"].sudo()
        report = self.env.ref("hr_expense_meal_allowance.action_report_hr_expense_meal_allowance").sudo()
        print_report_name = report.attachment or report.print_report_name
        for expense in self.filtered("is_meal_allowance"):
            pdf_content, _mime = (
                report._render_qweb_pdf(report, expense.ids)
            )
            if report.attachment:
                continue
            if not print_report_name:
                attachment_name = f"{expense.name}.pdf".replace("/", "_")
            else:
                attachment_name = safe_eval(print_report_name, {"object": expense, "time": time})
            attachment_model.create(
                {
                    "name": attachment_name,
                    "raw": pdf_content,
                    "res_model": "hr.expense",
                    "res_id": expense.id,
                    "mimetype": "application/pdf",
                }
            )
```

hr_expense_meal_allowance/report/expense_included_meal.xml
```diff
@@ -0,0 +1,197 @@
...
+                    <div class="col-9 text-muted">
+                        <span t-field="o.customer_id">Azure Interior</span>
+                    </div>
+                </div>
+                <div class="row">
+                    <div class="col-3 fw-bold">
+                        <span>Whole Days:</span>
+                    </div>
+                    <div class="col-9 text-muted">
+                        <span t-field="o.number_of_travel_days">1</span>
```
mt-software-de:
```suggestion
                        <span t-field="o.number_of_days">1</span>
```

hr_expense_meal_allowance/report/expense_included_meal.xml
```diff
@@ -0,0 +1,197 @@
...
+                    </div>
+                </t>
+            </t>
+        </t>
+    </template>
+    <template id="document_expense_meal_allowance">
+        <div class="page" t-if="o.is_meal_allowance">
+            <div class="oe_structure" id="oe_structure_meal_allowance_section" />
+            <p style="page-break-after:always;" />
+            <h2 style="display: inline-block;">
```
mt-software-de, line 87: Wouldn't it make sense to move the page-break? 
Otherwise each `Meal allowance Report` will start with a blank page. 
I guess thats want @smaddlsoft mentioned. 

```suggestion
        <xpath expr="//div[hasclass('page', 'o_content_pdf')]" position="after">
            <p style="page-break-after:always;" />
            <t t-call="hr_expense_meal_allowance.document_expense_meal_allowance" />
        </xpath>
    </template>
    <template id="report_meal_allowance">
        <t t-call="web.html_container">
            <t t-foreach="docs" t-as="o">
                <t t-call="web.external_layout">
                    <div class="page o_content_pdf">
                        <div
                            class="oe_structure"
                            id="oe_structure_report_meal_allowance"
                        />
                        <t
                            t-call="hr_expense_meal_allowance.document_expense_meal_allowance"
                        />
                    </div>
                </t>
            </t>
        </t>
    </template>
    <template id="document_expense_meal_allowance">
        <div class="page" t-if="o.is_meal_allowance">
            <div class="oe_structure" id="oe_structure_meal_allowance_section" />
            <h2 style="display: inline-block;">
```

mt-software-de (changes requested):

hr_expense_meal_allowance/models/hr_expense.py
CRogos: Claude short answer: negligible CPU difference, but they are **not** equivalent in one way that can matter — the prefetch set.

**What's the same**

`filtered` with a string resolves to the field descriptor itself ([models.py](odoo/odoo/orm/models.py) — `func = self._fields[func].__get__`), so it's the exact same field read as `expense.is_meal_allowance`. Both trigger one batched compute of `_compute_meal_allowance_tag` over the whole recordset. `filtered` costs one extra Python pass plus one tuple/recordset allocation — microseconds, irrelevant next to the `search_count` per record in that loop.

**What's different**

`filtered` returns `self.browse(ids)`, and `browse` sets `_prefetch_ids = ids` (the *filtered* ids). The `continue` version keeps iterating singletons whose `_prefetch_ids` is the **full** `self`.

So in [hr_expense.py:245](user/modules/oca/l10n-germany/hr_expense_meal_allowance/models/hr_expense.py#L245), the first `expense.name` read:
- `filtered` → one query fetching only the meal-allowance expenses
- `continue` → one query fetching *all* expenses in `self`, including the ones you skipped

If 5 of 500 expenses are meal allowances, that's a 100× smaller read. `filtered` wins here, and it's the version you already have.

**The one case where it flips**

A narrower prefetch set is a pessimization if the *same* fields are later read on the records you filtered out — you get two queries instead of one. That's what happens at [hr_expense.py:176-179](user/modules/oca/l10n-germany/hr_expense_meal_allowance/models/hr_expense.py#L176-L179), where you split `self` into both halves with two `filtered(lambda ...)` calls: each half gets its own prefetch set, so fields read on both halves are fetched twice. `grouped("is_meal_allowance")` is the tool for that split — it explicitly passes `prefetch_ids=self._prefetch_ids`, keeping both halves in one prefetch set. (Also, the lambda form there is slower than the string form, since it can't use the descriptor shortcut.)

Verdict: keep `filtered("is_meal_allowance")` at line 245. Prefer `filtered` generally — it's faster in the common "skip most records" case and clearer. Only reach for the `continue` form when you need the loop variable to stay in the original prefetch set.

hr_expense_meal_allowance/report/expense_included_meal.xml
CRogos: Basically I agree, but improved it further.

hr_expense_meal_allowance/models/hr_expense.py
mt-software-de: The prefetch argument doesn't apply here: `filtered("is_meal_allowance")` computes the field over all of self, which reads product_id and so already loads all stored columns for the whole recordset. The narrower prefetch set of the filtered result saves nothing. Either form is fine; no strong opinion.

PR #231, opened 2026-04-24
[18.0] [ADD] l10n_de_invoice_labor_cost

ollo2012:
Adds calculation and Note for labor cost products on invoice report


tv-openbig:
@ollo2012 Maybe i didn't get it, but what is the use case for a german company ? Isn't it related to private persons tax declaration ? Best regards and thanks for this PR.

OSevangelist:
@tv-openbig indeed, but a craftman needs to have a special paragraph on the invoice to allow the inividual end-user to deduct it from his / her tax statement. This module just implements that "German" resquirement.

ollo2012:
exactly @OSevangelist @tv-openbig Some People add it as note on there invoices. This module is meant to do this automatically based on the product configuration. It's an actual requirement from @MrTango

MrTango:
Yes, when you have an invoice for B2C, you need to have this extra part which says that the invioces contains:
Im Rechnungsbetrag von 1.500,00 EUR sind 900,00 EUR Lohnkosten enthalten. Darin enthaltene Arbeitskosten (netto): 756,30 EUR Darin enthaltene Umsatzsteuer: 143,70 EUR

pniederlag (approved):
LGTM

l10n_de_invoice_labor_cost/models/account_move.py
```diff
@@ -0,0 +1,41 @@
+# Copyright 2025 Maik Derstappen (https://derico.de)
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import api, fields, models
+
+
+class AccountMove(models.Model):
+    _inherit = "account.move"
+
+    l10n_de_labor_cost_net = fields.Monetary(
```
mt-software-de: IMO we should stick to more odoo like naming

```suggestion
    l10n_de_labor_cost_untaxed = fields.Monetary(
```

l10n_de_invoice_labor_cost/models/account_move.py
```diff
@@ -0,0 +1,41 @@
...
+        string="Labor Cost (Net)",
+        compute="_compute_labor_cost_values",
+        help="Total net amount of labor cost lines",
+    )
+    l10n_de_labor_cost_tax = fields.Monetary(
+        string="Labor Cost (Tax)",
+        compute="_compute_labor_cost_values",
+        help="Total tax amount on labor cost lines",
+    )
+    l10n_de_labor_cost_gross = fields.Monetary(
```
mt-software-de: Odoo like naming. 
```suggestion
    l10n_de_labor_cost_total = fields.Monetary(
```

l10n_de_invoice_labor_cost/models/account_move.py
```diff
@@ -0,0 +1,41 @@
...
+        string="Labor Cost (Gross)",
+        compute="_compute_labor_cost_values",
+        help="Total gross amount of labor cost lines (net + tax)",
+    )
```
mt-software-de: Add `store=True` reason explained in the other thread. 


```suggestion
    l10n_de_labor_cost_net = fields.Monetary(
        string="Labor Cost (Net)",
        compute="_compute_labor_cost_values",
        help="Total net amount of labor cost lines",
        store=True,
    )
    l10n_de_labor_cost_tax = fields.Monetary(
        string="Labor Cost (Tax)",
        compute="_compute_labor_cost_values",
        help="Total tax amount on labor cost lines",
        store=True,
    )
    l10n_de_labor_cost_gross = fields.Monetary(
        string="Labor Cost (Gross)",
        compute="_compute_labor_cost_values",
        help="Total gross amount of labor cost lines (net + tax)",
        store=True,
    )
```

l10n_de_invoice_labor_cost/models/account_move.py
```diff
@@ -0,0 +1,41 @@
...
+        for move in self:
+            labor_lines = move.invoice_line_ids.filtered(
+                lambda line: line.product_id.is_labor_cost_product
+            )
+
+            move.l10n_de_labor_cost_net = sum(labor_lines.mapped("price_subtotal"))
+            move.l10n_de_labor_cost_tax = sum(
+                line.price_total - line.price_subtotal for line in labor_lines
+            )
+            move.l10n_de_labor_cost_gross = sum(labor_lines.mapped("price_total"))
```
mt-software-de: Hey thx. But looks not quite correct. :-)

We need to ensure that the labor costs of a posted move will not change. This isn't the case with your implementation. 

You will run into an error if you have an posted move, which once was calculated with labor costs and the flag `is_laber_cost_product` of a product is changed.

If you have a move which is posted and it includes a laber cost product the computation will show you labor costs. Once you change the the flag on the product it will not show you labor costs. For a posted move this is not good. But this could be easily fixed, by making the fields storable. And not including `is_labor_cost_product` in the depends of the method.
Also to ensure the labor costs is correctly computed you should add `state` in the depends.

mt-software-de: Here some technical code review. 

You go through the list several places. This is not needed. And shouldn't be done. 
Even if it has not a big performance impact. 

1. filter is going through the list
2. mapped price_subtotal is going through the filtered list
3. calculating tax `for line in labor_lines` is going through the filtered list
4. mapped price_total is going through the filtered list


```suggestion
            cost_untaxed = 0
            cost_total = 0
            for line in move.invoice_line_ids:
                if not line.product_id.is_labor_cost_product:
                    continue
                cost_untaxed += line.price_subtotal
                cost_total += line.price_total
            move.l10n_de_labor_cost_untaxed = cost_untaxed
            move.l10n_de_labor_cost_total = cost_total
            move.l10n_de_labor_cost_tax = costs_total - cost_untaxed
```

Just minor,
IMO the logic should move into account.move.line. 
But just because you have more flexibility later if you would need do some fancy stuff.
So nothing for now.

mt-software-de (changes requested):
Thx.

l10n_de_invoice_labor_cost/models/account_move.py
```diff
@@ -0,0 +1,41 @@
...
+        help="Total gross amount of labor cost lines (net + tax)",
+    )
+
+    @api.depends(
+        "invoice_line_ids.tax_ids",
+        "invoice_line_ids.price_subtotal",
+        "invoice_line_ids.price_total",
+    )
+    def _compute_labor_cost_values(self):
+        for move in self:
```
mt-software-de, line 36: Also i guess it only should be done for move_types out_invoice, out_refund?
```suggestion
        "invoice_line_ids.price_total",
        "move_type",
    )
    def _compute_labor_cost_values(self):
        for move in self:
            if move.move_type not in  ["out_invoice", "out_refund"]:
                move.l10n_de_labor_cost_tax = 0
                move.l10n_de_labor_cost_untaxed = 0
                move.l10n_de_labor_cost_total = 0
                continue
```

mt-software-de (changes requested):

l10n_de_invoice_labor_cost/models/account_move.py
mt-software-de: This should also be represented in a unittest.

smaddlsoft:
@ollo2012 @MrTango do you think you can get these changes done so we can reapprove and merge?

MrTango:
hey I'm sorry didn't get to it, but I will. I'm off next week, so will probably take me another couple if weeks.

pniederlag:
@MrTango drop us a note when you need some advice, help or a review.

l10n_de_invoice_labor_cost/models/account_move.py
ollo2012: Hey @mt-software-de 
I don't exactly understand what you mean by adding ""state in the depends"
Updated the PR. Maybe you can recheck!

ollo2012:
Updated the PR. Review and Feedback welcome.

l10n_de_invoice_labor_cost/models/account_move.py
mt-software-de: we should add `@api.depends("state")` to ` _compute_labor_cost_values` otherwise its maybe not correctly calculated if you posting a invoices

l10n_de_invoice_labor_cost/models/__init__.py
```diff
@@ -0,0 +1,3 @@
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from . import account_move, product_template
```
mt-software-de, line 3: Normally a line per file. 
```suggestion
from . import account_move
from . import product_template
```

l10n_de_invoice_labor_cost/models/account_move.py
```diff
@@ -0,0 +1,51 @@
...
+        compute="_compute_labor_cost_values",
+        help="Total gross amount of labor cost lines (net + tax)",
+        store=True,
+    )
+
+    @api.depends(
+        "invoice_line_ids.tax_ids",
+        "invoice_line_ids.price_subtotal",
+        "invoice_line_ids.price_total",
+        "move_type",
```
mt-software-de, line 33: To ensure it is correctly calculated when posting. 
```suggestion
        "move_type",
        "state",
```

l10n_de_invoice_labor_cost/models/product_template.py
```diff
@@ -0,0 +1,29 @@
...
+    is_labor_cost_product = fields.Boolean(
+        help="Indicates whether this product is a labor product.",
+    )
+
+    def _check_is_labor_cost_product(self):
+        self.ensure_one()
+        if self.is_labor_cost_product and self.type in ("consu", "combo"):
+            raise ValidationError(
+                _("Labor cost products cannot be of type 'Goods' or 'Combo'.")
+            )
```
mt-software-de, line 20: I think it is better to explicitly check the value we want to have.
If there is change in the selection values like it was done by adding combo in v18,
this will still work. 
```suggestion
        if self.is_labor_cost_product and self.type != "service":
            raise ValidationError(
                _("Labor cost products must be of type 'Service'.")
            )
```

l10n_de_invoice_labor_cost/models/account_move.py
```diff
@@ -0,0 +1,51 @@
+# Copyright 2025 Maik Derstappen (https://derico.de)
```
mt-software-de, line 1: Why isn't there your Copyright? Instead of Maik's?

l10n_de_invoice_labor_cost/readme/DESCRIPTION.md
```diff
@@ -0,0 +1,42 @@
...
+  - Net labor costs (labor)
+  - Included VAT
+
+## Technical Details
+
+- **Module Name**: `l10n_de_invoice_labor_cost`
+- **Category**: Localization/Germany
+- **License**: AGPL-3
+- **Version**: 18.0.1.0.0
+- **Dependencies**: `account`
```
mt-software-de, line 37: Is this necessary? You will find those information in the manifest.
For example version needs to be changed if there is a change in the module. 
There is an oca automation which is automatically changing the version in the __manifest__.py but not here so we will run into different versions here and in the manifest. 
```suggestion
```

l10n_de_invoice_labor_cost/__manifest__.py
```diff
@@ -0,0 +1,18 @@
...
+    "category": "Localization",
+    "license": "AGPL-3",
+    "author": "Ole Jancke, Odoo Community Association (OCA)",
+    "website": "https://github.com/OCA/l10n-germany",
+    "depends": ["account"],
+    "data": [
+        "views/product_template_views.xml",
+        "report/report_invoice.xml",
+    ],
+    "installable": True,
```
mt-software-de, line 17: This is a default value, no need to set it here. 
```suggestion
```

mt-software-de (changes requested):

smaddlsoft:
@ollo2012 do you wann do the final adjustments so we can merge?

PR #232, opened 2026-04-24
[19.0][ADD] account_move_date_visible: make field visible

pniederlag:
* as decided during code sprint


tv-openbig:
@pniederlag I have reviewed it. Very good idea to have that module available.

tv-openbig:
@mt-software-de Hello Michael, would you please review it as a 2nd reviewer in order to merge that module.

mt-software-de:
Yes i will do a review. 

@pniederlag could you please update the PR Title according to the oca conventions?

grafik

account_move_date_visible/__manifest__.py
```diff
@@ -0,0 +1,13 @@
+{
+    "name": "Show Invoice Dates (GoB)",
```
mt-software-de, line 2: IMO This should be renamed. account_move_date_show. 
No matter how the name of directory and the name in the __manifest__.py should be aligned.

But the real major change request, would be not to contribute it to this repo.
Because I think this addon is quite generic and could be used by more then just l10n_de.
So i would contribute it to https://github.com/OCA/account-invoicing
This would present the addon to a larger group of people, perhaps leading to its use by non-l10n_de users and possibly leads to more people maintaining it. 

To describe our german needs for this addon, we can then use the planned german "Must Have Accounting Addons" Addon. To describe how to use it in a german context, like it is done here with the Description.

mt-software-de (changes requested):

account_move_date_visible/__manifest__.py
pniederlag: Thx for your feedback. I'll fix it and try to shift it over to account-invoicing.

smaddlsoft: @pniederlag do you think we could get this over to account-invoicing?

account_move_date_show/__manifest__.py
```diff
@@ -0,0 +1,13 @@
+{
+    "name": "Show Invoice Dates (GoB)",
```
smaddlsoft: @pniederlag whats the status here? Would be nice to get this online, either here or better at account-invoicing

mt-software-de:
@pniederlag sorry i accidentally closed this PR by changing the base to v19.

PR #233, opened 2026-05-04
[ADD] l10n_de_account: rename menu 'Invoicing' -> 'Accounting'

yweng8111:
Hi,
I have added a new module that changes the main menu and the "Invoicing" settings menu to "Accounting". This feature is also implemented in the module account_usability (in oca/account-financial-tools). However, we were unable to add the German translation for it. Therefore, I fixed it using the same approach as in oca/l10n-italy (see pull request
https://github.com/OCA/l10n-italy/pull/4298)

Other pull requests related to this feature in account-financial-tools can be found here:
https://github.com/OCA/account-financial-tools/pull/2243
https://github.com/OCA/account-financial-tools/pull/2249

### DIFF
--- a/l10n_de_account/README.rst
+++ b/l10n_de_account/README.rst
@@ -1,0 +1,38 @@
+Usage
+=====
+
+Activate Translation of German
+
+
+Bug Tracker
+===========
+
+Bugs are tracked on `GitHub Issues
+<https://github.com/OCA/hr/issues>`_. In case of trouble, please
+check there if your issue has already been reported. If you spotted it first,
+help us smash it by providing detailed and welcomed feedback.
+
+Credits
+=======
+
+Contributors
+------------
+
+* Yu Weng <yweng@elegosoft.com>
+
+Do not contact contributors directly about support or help with technical issues.
+
+Maintainer
+----------
+
+.. image:: https://odoo-community.org/logo.png
+   :alt: Odoo Community Association
+   :target: https://odoo-community.org
+
+This module is maintained by the OCA.
+
+OCA, or the Odoo Community Association, is a nonprofit organization whose
+mission is to support the collaborative development of Odoo features and
+promote its widespread use.
+
+To contribute to this module, please visit https://odoo-community.org.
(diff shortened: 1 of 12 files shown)


smaddlsoft (approved):
Looks good to me (see screenshots), thanks @yweng8111 !

Will it conflict/also work when the account_usability module is installed that was so far handling the translation of the main menu item in english (but not German), or will the two work well alongside each other which they should? 

Image

Image

Image

Image

smaddlsoft:
@mt-software-de maybe you wanna review as well, potentially with the suggested test when account usability is installed as well, then @pniederlag could merge afterwards

mt-software-de:
Thx for doing it.
But IMO this is not a good approach. 
I don't like the fact that every localization needs to do this, as well that we can not use the default process for translating terms. And also to list the terms not in a *.po file. 
You can check my approach here https://github.com/OCA/account-financial-tools/pull/2287

mt-software-de (changes requested):
Thx for doing it.
But IMO this is not a good approach.
I don't like the fact that every localization needs to do this, as well that we can not use the default process for translating terms. And also to list the terms not in a *.po file.
You can check my approach here https://github.com/OCA/account-financial-tools/pull/2287

OSevangelist:
Hi @yweng8111 @mt-software-de if we all agree on the more general approach of @mt-software-de who can take responsibility to add it to account-financial-tools and close this PR?

mt-software-de:
> Hi @yweng8111 @mt-software-de if we all agree on the more general approach of @mt-software-de who can take responsibility to add it to account-financial-tools and close this PR?

The PR for it is already created. https://github.com/OCA/account-financial-tools/pull/2287

Do we all agree on it? @smaddlsoft
If so i think Reviews on the new PR would be great.

smaddlsoft:
I cant judge the best technical approach really, but what you describe as a generic approach sounds much better and that weblate can be used. So I prefer that way as well and will review the account_usability PR.

Thanks for this improvement @mt-software-de 

@yweng8111 do you wanna close this PR then?

yweng8111:
@smaddlsoft @OSevangelist Sorry, I've forgot to close it here. I prefer the solution from @mt-software-de .

PR #236, opened 2026-05-15
[Draft][ADD][18.0] DATEV API

fkantelberg:
After starting it in the code sprint in Hamburg I finally have a function PoC with working OAuth authorization, long term tokens, and a revocation process (required by DATEV).

This PR introduces the DATEV API via a new module `datev_api` and also offers a module to upload the files from `datev_api_export_dtvf` directly to the API as first test balloon. It's still work in progress at this point as there are some open questions:

- Refactoring of `datev_export` might be needed because this API module requires the client and consultant number but if you only want to use the API to import DTVF it's weird to install `datev_export`
- Refactoring of `datev_export_dtvf` might be needed to not deconstruct the zip files for the upload. Also partial upload isn't handled well (e.g. if only the Buchungsstapel could be uploaded but not the other files).
- At this point I'm not sure where to check if the upload succeeded in the DATEV portal
- Refactoring of `datev_export_xml` as the used invoice XML format isn't supported by the API and only [ledger XML](https://developer.datev.de/de/file-format/details/datev-xml-interface-online/format-specification-) is
- I haven't checked the importing side but it should work similar to the example


fkantelberg:
**How to test:**
- Login/Register at https://developer.datev.de
- Create organisation with a valid consultant number
- Create app
  - `OpenID Connect Authorization Code Flow` as Authorization Flow
  - `Confidential` if possible as Client Type

After creation of the App the App credentials has to be set in Odoo as Client ID and Client Secret. During the authorization you need to use the Sandbox credentials to login and accept the access. With the current modules it's recommended to subscribe to `accounting:ext-files` (datev_api_export_dtvf) and `accounting:clients` (datev_api)

smaddlsoft:
@fkantelberg 

Thanks a lot for this! Much expected and useful for the community.

I tried testing it, some questions:
- I ran into Error 500 after entering API credentials in Odoo, with API Mode "Testing", going on "Authenticate", chosing "Datev Benutzerkonto" as Login Method and then using the Sandbox User and PW for this login.

- I am testing with a freshly created Datev Developer account user in a freshly created organisation in Datev, with no actual existing Datev client number. Will this work for testing, if so, how?

- How is the intended "flow" for "Odoo/Datev endusers" ? They signup themselves on Developer portal etc.? Their Odoo partners do this for them? Their tax advisors? As I couldnt test the full flow and functionality, this is a bit unclear to me right now. 

Thanks in advance

fkantelberg:
@smaddlsoft
1) 500 on the redirect? Do you have any traceback here to look into?

2) There are some clients automatically created in the sandbox with different possible services. I guess I used the client number 1 for me which had the most access. There is an endpoint to fetch those but I think it's only necessary for testing. (similar to what `_datev_client_information` does but without passing the `client_id` in the url)

3) [FAQ](https://developer.datev.de/de/help) states that DATEV recommends that the company has the ownership of the API things which means that the company has to create the organisation and account on developer.datev.de

smaddlsoft:
@fkantelberg happy to try testing again with your support. 
Related question: Do you think we could finally add inclusion of analytical accounts into the Datev export? If so, where/how to best open and describe an issue for that so that potentially both the API and the existing DTVF and/or XML export could benefit from it?

fkantelberg:
I only see some fields (KOST1, KOST2, KOST) in the DTVF [1] and XML ledger [2] file. KOST1 is getting filled already. I currently don't see a way to synchronize the analytic accounts directly besides the booking batches. 

[1] https://developer.datev.de/en/file-format/details/datev-format/format-description/booking-batch

[2] https://developer.datev.de/en/file-format/details/datev-xml-interface-online/format-specification-/record-data-file-

smaddlsoft:
> I only see some fields (KOST1, KOST2, KOST) in the DTVF [1] and XML ledger [2] file. KOST1 is getting filled already. I currently don't see a way to synchronize the analytic accounts directly besides the booking batches.
> 
> [1] https://developer.datev.de/en/file-format/details/datev-format/format-description/booking-batch
> 
> [2] https://developer.datev.de/en/file-format/details/datev-xml-interface-online/format-specification-/record-data-file-

Cool, I wasnt aware and will test that.
@smaddlsoft

smaddlsoft:
> > I only see some fields (KOST1, KOST2, KOST) in the DTVF [1] and XML ledger [2] file. KOST1 is getting filled already. I currently don't see a way to synchronize the analytic accounts directly besides the booking batches.
> > [1] https://developer.datev.de/en/file-format/details/datev-format/format-description/booking-batch
> > [2] https://developer.datev.de/en/file-format/details/datev-xml-interface-online/format-specification-/record-data-file-
> 
> Cool, I wasnt aware and will test that. @smaddlsoft

@fkantelberg you were right. Its included/handled in both DTVF and XML exports. Maybe we should update the readme of the modules to make that clearer. Will note this in our roadmap.

fkantelberg:
@smaddlsoft I tested it locally again and it works flawlessly. I only selected SKR04 before. This shouldn't matter at this state but for transparency. Minimally the DATEV product `accounting:extf-files` is only required for the current modules. More shouldn't hurt though. 500 is on DATEV side please make sure to not include whitespaces in the client secret and client ID. I hope that helps.

smaddlsoft:
@mt-software-de can you get this to work?

I tried multiple times, always the same error on Datev side.
Have tried on Runboat though, will try it on a dedicated system.

image

fkantelberg:
@smaddlsoft Can you try a chromium based browser? I read in [1] that they don't recommend firefox. If I use firefox I also see the 500 locally but I don't know where this comes from or what DATEV did there.

[1] https://wissensplattform.apps.datev.de/help/document/1038725

smaddlsoft:
@fkantelberg I have exclusively tested with Chrome..

smaddlsoft:
@fkantelberg let me check again once I have a valid system up and running. Might actually be the redirect after all, and in runboat due to http I cant set the actual redirect URL, doesnt accept anything but localhost in Datev for http, which might explain why its working for you.

See here: "the app registered in the DATEV Developer Portal must contain the exact redirect URL https://<your-domain>/web/datev/oauth/callback. Production redirect URLs must be HTTPS — localhost, IP addresses and custom schemes are not allowed and lead to the app being blocked by DATEV (since 2026-03-01)."

https://github.com/aquarius-ventures/odoo-datev-connector

smaddlsoft:
@fkantelberg PS: tried Edge, same 500 error.

fkantelberg:
You can easily switch to `https` in the runboat (but have to do it manually). The certificate warning can be ignored. It could be because I configured localhost also as redirect URL in DATEV.

PR #237, opened 2026-05-16
[ADD] l10n_de_accounting_app: German accounting setup

mt-software-de:
This addon gives German Odoo users a single, guided starting point for setting up their accounting.

Running accounting in Germany usually means assembling a handful of OCA modules. Knowing *which* modules to pick, and where they live, is a hurdle for newcomers.

This addon collects that knowledge into one place. It adds a dedicated **German Accounting** page in *Settings* where an accounting manager can enable the recommended add-ons, organised by topic. Each option comes with a short explanation and a link to the OCA repository it comes from, so it doubles as a guided tour of the German accounting ecosystem.

The curated list of add-ons was discussed and agreed on at the German OCA sprint in Hamburg in April 2026.

Because these add-ons are deliberately optional (not hard dependencies), the page also checks whether a chosen module is actually available in the installation. If it isn't, the option is unticked and the user gets a clear hint to add the relevant OCA repository and update the apps list — instead of a checkbox that silently does nothing on save.

Under the hood, the same information about each add-on lives in three places: the settings model (the fields), the settings view (the UI), and the module's documentation. Keeping those in sync by hand is tedious and error-prone — it's easy to add a module to the screen but forget the README, or to let a description drift between the UI and the docs. To avoid that, all three are generated from a single `config.yml` file that lists the sections and add-ons. To add or change an entry you edit one place and regenerate; the model, the view, and the documentation always stay consistent, and contributing a new module is a small, obvious change instead of a hunt through several files.


mt-software-de:
@smaddlsoft sorry that it took so long. 
Please have a first look. 

Here you can see how the README looks like. 
https://github.com/mt-software-de/l10n-germany/tree/18-add-l10n_de_accounting_app/l10n_de_accounting_app

smaddlsoft (changes requested):
@mt-software-de Many thanks! No worries for the timing. 

Very nice to have it in the config.

image

I tried uninstalling a few modules via the config settings, and it seemed to work fine and give warnings before uninstalling also. Installing I could only test those which came preinstalled on runboat as I suppose the other repos are not available in the runboat instance? 

So from my side, functional review, looks good overall, with these observations/ questions/change requests for now:
1) Should there be some kind of warning/UX info that a module is not available/installable but can be obtained from [github repo link]?
2) Why did you include the UBL/CII given this is a Odoo core module activated in german localisation by default?
3) Whats the rational for the choice of modules that come activated/installed with this module? Should we select all/some/none, and what was the advantage/disadvatage again?

Thanks again!

smaddlsoft:
@pniederlag @fkantelberg @tv-openbig would also appreciate your input/feedback on this one.

smaddlsoft:
PS: once ready/merged, we need to add [Datev API](https://github.com/OCA/l10n-germany/pull/236) to this module

mt-software-de:
> @mt-software-de Many thanks! No worries for the timing.
> 
> Very nice to have it in the config.
> image
> 
> I tried uninstalling a few modules via the config settings, and it seemed to work fine and give warnings before uninstalling also. Installing I could only test those which came preinstalled on runboat as I suppose the other repos are not available in the runboat instance?
> 
> So from my side, functional review, looks good overall, with these observations/ questions/change requests for now:
> 
>     1. Should there be some kind of warning/UX info that a module is not available/installable but can be obtained from [github repo link]?
> 
>     2. Why did you include the UBL/CII given this is a Odoo core module activated in german localisation by default?
> 
>     3. Whats the rational for the choice of modules that come activated/installed with this module? Should we select all/some/none, and what was the advantage/disadvatage again?
> 
> 
> Thanks again!

1. I check what i can do.
2. It was on my list created on the sprint, but this is not a dependency of odoo's l10n_de addon in v18.
3. This addon does not install by default any addon, if one is already installed it is because of the runboat installation. And this shows also in the res.config.settings view as activated.

smaddlsoft:
@mt-software-de 
1) cool!
2) should we remove it then? Find it counterintuitive/confusing.
3) Understood from this module point of view, though not why runboat comes with these modules preinstalled as usually it comes blank in my understanding apart from the dependencies of the module tested in runboat.

mt-software-de:
> @mt-software-de
> 
>     1. cool!
> 
>     2. should we remove it then? Find it counterintuitive/confusing.
> 
>     3. Understood from this module point of view, though not why runboat comes with these modules preinstalled as usually it comes blank in my understanding apart from the dependencies of the module tested in runboat.

2. yes but IMO for v19 this v18

mt-software-de:
> @mt-software-de
> 
>     1. cool!
> 
>     2. should we remove it then? Find it counterintuitive/confusing.
> 
>     3. Understood from this module point of view, though not why runboat comes with these modules preinstalled as usually it comes blank in my understanding apart from the dependencies of the module tested in runboat.

3. Not runboat install a addons of the repo. Including as its dependencies. 
I checked it only addons of this repo listed as installed/checked

smaddlsoft:
2. Why in v19 not here?

3. Okay, thanks, learned something new.

l10n_de_accounting_app/views/res_config_settings_views.xml
```diff
@@ -0,0 +1,169 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+    <record id="res_config_settings_view_form" model="ir.ui.view">
+        <field name="name">res.config.settings.l10n.de.accounting.app.view</field>
+        <field name="model">res.config.settings</field>
+        <field name="priority" eval="100" />
+        <field name="inherit_id" ref="account.res_config_settings_view_form" />
+        <field name="arch" type="xml">
+            <xpath expr="//app[@name='account']" position="inside">
```
mt-software-de: @smaddlsoft instead of adding it to account settings page. 
Should we maybe create an own for it?

Image

mt-software-de:
> 2. Why in v19 not here?
> 
>     3. Okay, thanks, learned something new.

2. This is PR for version 18 -> Odoo v18.0 does not include it by default in l10n_de
Odoo version 19.0 includes it by default, when there is a migration PR for this addon to migrate to version 19 we can remove it.

l10n_de_accounting_app/views/res_config_settings_views.xml
smaddlsoft: Nice idea. If its clearly indented and thus a submenu to invoicing: why not? makes things clearer/more seperated. Technically not familiar with any pros and cons of this approach.

smaddlsoft:
> 2. This is PR for version 18 -> Odoo v18.0 does not include it by default in l10n_de
>    Odoo version 19.0 includes it by default, when there is a migration PR for this addon to migrate to version 19 we can remove it.

Are we talking account_edi_ubl_cii? This gets installed in all my v18 runbot and runboat tests of Odoo community, even without German company/localisation.

image

l10n_de_accounting_app/views/res_config_settings_views.xml
fkantelberg: Not sure if a new application is actually required here because if the installed modules are implementing additional configurations they are usually also visible in `Invoicing` again and you would have to go back and forth.

On the other side most of the `blocks` here have only 1 `setting` for only 1 boolean which makes everything really sparse. I don't know if it's worth to put everything in 1 `<block>` and use `<div class="content-group">` to group them.

l10n_de_accounting_app/readme/DESCRIPTION.md
```diff
@@ -0,0 +1,86 @@
...
+
+- **l10n_de_mis_reports** ([OCA/l10n-germany](https://github.com/OCA/l10n-germany)) — Profit/Loss, Balance Sheet, SKR03, SKR04.
+
+## Bank Import
+
+From [OCA/bank-statement-import](https://github.com/OCA/bank-statement-import):
+
+- **account_statement_import_camt54** — CAMT.054 schema.
+- **account_statement_import_online_paypal** — PayPal.
+- **account_statement_import_online_ponto** — Ponto API, fetches statements automatically.
```
fkantelberg, line 45: Can you explain why you choose those? I know customers which are using `stripe` as import too.

l10n_de_accounting_app/views/res_config_settings_views.xml
mt-software-de: @fkantelberg what do you mean with a new application? The addon itself or the just the view on res.config.setting?

mt-software-de: > Nice idea. If its clearly indented and thus a submenu to invoicing: why not? makes things clearer/more seperated. Technically not familiar with any pros and cons of this approach.

I will create for now an own dedicated page.

l10n_de_accounting_app/readme/DESCRIPTION.md
mt-software-de: We just discussed on the sprint, and that was the choice.
We wanted to have one recommendation. 
But IMO we can add as well stripe here.

mt-software-de:
> > 2. This is PR for version 18 -> Odoo v18.0 does not include it by default in l10n_de
> >    Odoo version 19.0 includes it by default, when there is a migration PR for this addon to migrate to version 19 we can remove it.
> 
> Are we talking account_edi_ubl_cii? This gets installed in all my v18 runbot and runboat tests of Odoo community, even without German company/localisation.

Yes in runbot and runboat, because runbot (odoo) installs more or less all addons, runboat installs all addons of the repo and its dependencies. But which a fresh empty database. It will be not installed.

l10n_de_accounting_app/views/res_config_settings_views.xml
fkantelberg: I meant the `<app>` and the dedicated page with that.

l10n_de_accounting_app/readme/DESCRIPTION.md
mt-software-de: @smaddlsoft ?

l10n_de_accounting_app/views/res_config_settings_views.xml
mt-software-de: Ok. IMO a dedicated app is what we need. :-)

mt-software-de:
The PR is now ready.
I added the logic to check if an addon is available which was ticked.
I created an own page for it in the config settings view. 
And added the list of addons and description into one config.yml file, to make it easier to maintain, this config.yml is the used to extend the model, view and readme/description

@smaddlsoft

smaddlsoft (approved):
Looks good to me from a functional perspective.

Small thing: if possible, as said previously, I would indent the "German accounting" in the settings menu sidebar, so it doesnt break (too much) the logic that each main app that is visible on the dashboard also has one setting page, and nothing else. Alternatively maybe do away with the app and keep in a section under Accounting settings as before.

l10n_de_accounting_app/readme/DESCRIPTION.md
pniederlag: This will alwyas be an opnininated list. I dont see a reason to leave out stripe. It could be added anytime but i wouldn't want to block this PR if it's still missing.

pniederlag (approved):
LGTM

pniederlag:
@fkantelberg any objectons from your side against merging?

mt-software-de:
> @fkantelberg any objectons from your side against merging?

Wait until the Meeting today.

mt-software-de:
@smaddlsoft i added you as well as maintainer, is this ok for you?

pniederlag:
- waiting for addition of paypal
- waiting for feedback from @fkantelberg on the dev script

l10n_de_accounting_app/readme/DESCRIPTION.md
mt-software-de: as discussed right now in the monthly meeting, we will add stripe as well.

l10n_de_accounting_app/dev_tools/generate.py
```diff
@@ -0,0 +1,236 @@
...
+    MODEL_PATH.write_text(render_model(cfg))
+    VIEW_PATH.write_text(render_view(cfg))
+    DESCRIPTION_PATH.write_text(render_description(cfg))
+    print(f"Wrote {MODEL_PATH.relative_to(ROOT)}")
+    print(f"Wrote {VIEW_PATH.relative_to(ROOT)}")
+    print(f"Wrote {DESCRIPTION_PATH.relative_to(ROOT)}")
+
+
+if __name__ == "__main__":
+    main()
```
fkantelberg, line 127: Looks ok. I would suggest to use templating instead of collecting formatted lines. Could it be an option to also use `copier` or `jinja2` here? Might help maintaining it.

fkantelberg (commented):
LGTM just one comment about the script to make it maybe better maintainable.

l10n_de_accounting_app/dev_tools/generate.py
mt-software-de: yes i thought so too.

mt-software-de: now it should be better.

fkantelberg (approved):

mt-software-de:
@pniederlag this is now ready to be merged.

PR #242, opened 2026-06-16
[18.0][FIX] datev_export_xml: update_attachment helper method

jans23:
The model has been renamed in Odoo 14.0.

### DIFF
--- a/datev_export_xml/tests/test_datev_export.py
+++ b/datev_export_xml/tests/test_datev_export.py
@@ -130,8 +130,10 @@ class TestDatevExport(BaseCommon):
 
             doc_file = "document.xml"
             inv_file = str(invoice + ".xml")
+            pdf_file = str(invoice + ".pdf")
             doc_data = z.read(doc_file)
             inv_data = z.read(inv_file)
+            pdf_data = z.read(pdf_file)
             # document.xml
             doc_root = etree.fromstring(doc_data)
             # invoice.xml file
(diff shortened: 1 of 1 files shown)


jans23:
@fkantelberg could you review this, please?

pniederlag:
@jans23 which problem/bug does this PR adress? Please add a description so we can check the problem you are facing, thx.

jans23:
@pniederlag I added a description as requested. Please have a look at the code. The model has been renamed in Odoo 14.0.

mt-software-de:
@mt-software-de please review

pniederlag:
@mt-software-de why is the test not failing? can this method be improved so it generates real value?

datev_export_xml/tests/test_datev_export.py
mt-software-de, line 334: The method can be removed completely, it is been used but the the returned attachment never. And there is not any tests which checks if the attachment is linked to the created invoice. It also doesn't make sense to test it all if it is linked or not, because it is just linked here in the test. So IMO its better to just remove this method.

mt-software-de (changes requested):

datev_export_xml/tests/test_datev_export.py
jans23: AFAIK we use this method otherwise we wouldn't have hit this error. @fkantelberg what do you think?

fkantelberg: The result of the function isn't important but the call is because of `find_existing_attachments` which will find the test attachment.

mt-software-de: But then a test should fail, otherwise the call is useless. @jans23 could you please extend the tests so that they are failing without your fix?

jans23: @mt-software-de done

fkantelberg (approved):

mt-software-de (approved):
Thx.

PR #246, opened 2026-07-06
[19.0][ADD] hr_expense_meal_allowance_trip

CRogos:


smaddlsoft:
@CRogos ready to review?

smaddlsoft (changes requested):
@CRogos 

Nice work. I positively tested creation of meal allowances & trips, starting from both sides.

Some things I noticed:
- The UX flow when creating an expense may be improved:
* when I first create an expense, then I see the trip field before I start filling the expense dates, which encourages me to create a trip, with its respective dates, but then these dates dont feed back into the expense dates. Additionally, when starting directly from the trip (which may be a natural flow), then it would be particularly nice if the dates would be pre-filled from the trip, which currently also doesnt happen. Maybe a dedicated button for meal allowance creation from trip could be an idea (as we talked about)
* its possible to add meal allowance date ranges outside the trip range. Maybe you want to give a blocking or informative (maybe preferred) warning.

Thanks!
Jan

Issue #228, opened 2026-04-24
Code Sprint Hamburg 2026


mt-software-de:
Here a list of Created/Closed/Worked PRs.

- [x] https://github.com/OCA/repo-maintainer-conf/pull/138

**Tax Statement (Umsatzsteuervoranmeldung)**

- [x] https://github.com/OCA/l10n-germany/pull/225

- [x] https://github.com/OCA/l10n-germany/pull/227

**Datev**

- [x] https://github.com/OCA/l10n-germany/pull/224

- [x] https://github.com/OCA/l10n-germany/pull/226

- [ ] https://github.com/OCA/l10n-germany/pull/236

**Labor Costs (Haushaltsnahe Dienstleistung)**

- [ ] https://github.com/OCA/l10n-germany/pull/231

**DE Finance Report**

- [x] https://github.com/OCA/l10n-germany/pull/229

- [ ] https://github.com/OCA/l10n-germany/pull/232

- [x] https://github.com/OCA/l10n-germany/pull/237

- [x] https://github.com/OCA/oca-github-bot/pull/346

- [ ] https://github.com/pledra/odoo-llm-base-ocr

muaazsiddiq:
- [x] l10n_de_mis_reports: #229

smaddlsoft:
-  [x] [18.0] #237

smaddlsoft:
- [ ] [18.0] #236

pniederlag:
@smaddlsoft @mt-software-de I'd like to get this issue solved. :-> Long-Term goals should be moved to our roadmap https://github.com/OCA/l10n-germany/pull/251 

Hope you didn't mind, I took the freedom do modifiy your comments by adding status checkboxes. Please let me know if you have objections on this.

smaddlsoft:
@pniederlag , good idea. Thanks, very nice with the checkboxes. 
have moved the open issues account_move_date_visible and the invoice_labor_cost also to the roadmap section on the de_accounting_app

smaddlsoft:
@pniederlag you can close the issue I suppose.
The LLM Base is not related and not very Germany specific.

pniederlag:
Remaining issues are tracked via roadmap. I consider the sprint a huge success, thx to everyone spending their time trying to improve open source software. Big kudos to our mentor @tv-openbig

PR #251, opened 2026-09-01
[18.0][IMP] l10n_de_accounting_app: add hr_expense_meal_allowance

smaddlsoft:
Adds `hr_expense_meal_allowance` to the German accounting app so it can be installed from the app's settings page, alongside the other curated OCA addons.

Handles meal allowances (`Verpflegungsmehraufwand`) as expenses with the German standard allowance rates per country and city.

As with the other addons in this module, the change is made in `config.yml` and the model, settings view and `readme/DESCRIPTION.md` are regenerated via `dev_tools/generate.py`.

---

Following review feedback from @mt-software-de and @pniederlag, the roadmap fragment that was originally part of this PR has been split out into a dedicated tracking issue, so this PR is now scoped to the meal allowance addon only.

### DIFF
--- a/l10n_de_accounting_app/config.yml
+++ b/l10n_de_accounting_app/config.yml
@@ -103,6 +103,16 @@ sections:
         label: Zusammenfassende Meldung (ZM)
         description: Extension for the Zusammenfassende Meldung (intra-community).
 
+  - name: Expense accounting (travel/meal)
+    repo: l10n-germany
+    description: null
+    addons:
+      - name: hr_expense_meal_allowance
+        label: Meal Allowance Expenses
+        description:
+          For handling meal allowances as expenses with the German standard allowance
+          rates per country and city.
+
   - name: Payment
     repo: bank-payment
     description: null
--- a/l10n_de_accounting_app/models/res_config_settings.py
+++ b/l10n_de_accounting_app/models/res_config_settings.py
@@ -17,6 +17,7 @@ _MODULE_FIELDS = [
     "module_datev_import_csv_dtvf",
     "module_l10n_de_tax_statement",
     "module_l10n_de_tax_statement_zm",
+    "module_hr_expense_meal_allowance",
     "module_account_payment_mode",
     "module_account_payment_order",
     "module_account_banking_sepa_credit_transfer",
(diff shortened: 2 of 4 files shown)


smaddlsoft:
@mt-software-de I thought better than a dedicated issue we use the existinng l10n_de_accounting_app roadmap for what we discussed today? Hope I opened I sugegsted this the right way. IF you have additional major suggestions for the roadmap, please add

smaddlsoft:
@mt-software-de hope I got it right finally using the generated files rather than manually editing.

smaddlsoft:
@pniederlag in case you wanna review though minor change but to follow protocol, before Michael merges

pniederlag:
@smaddlsoft in general it's propably preferred to have separate PR's per concern. yet for this PR it doesn't add that much value. @fkantelberg or @mt-software-de any comments from your side?

mt-software-de:
> @mt-software-de I thought better than a dedicated issue we use the existinng l10n_de_accounting_app roadmap for what we discussed today? Hope I opened I sugegsted this the right way. IF you have additional major suggestions for the roadmap, please add

Thanks for writing this up, @smaddlsoft - really useful overview.

But I don't think readme/ROADMAP.md is the right place for it. Most items belong to other addons and point to open PRs, so the content is outdated the moment those get merged or closed — and since the fragment ends up in the committed and published README.rst, the stale version stays there and gets carried along on every migration.
A tracking issue (follow-up to #228) fits much better: editable, checkable, no stale docs. ROADMAP.md could then just cover the limitations of the app module itself.

Also +1 to @pniederlag on splitting out hr_expense_meal_allowance into its own PR.

pniederlag:
thx @mt-software-de, very well valid points. @smaddlsoft we shouldn't make the roadmap static per version, How about adding named roadmap issues? probably something to brainstorm during oca days, unless we have a good idea upfront.

smaddlsoft:
@pniederlag @mt-software-de Understood and agreed. Trimmed this PR to the meal allowance addition then, kindly test/review and merge.

The Roadmap went to #253 

If you think its a good idea and necessary, we could link the roadmap-issue in the roadmap of the de_accounting_app

mt-software-de (approved):
LGTM

PR #254, opened 2026-09-07
[IMP/PORT][18.0] hr_expense_meal_allowance: treat start and end time of 00:00 as full days

CRogos:
Related to: https://github.com/OCA/l10n-germany/pull/223


CRogos:
I think the failed tests are not related to this change.

PR #255, opened 2026-09-11
[19.0][ADD] l10n_din5008_report_marks: new module

NICO-SOLUTIONS:
This module adds configurable DIN 5008 fold and punch marks to reports using the DIN 5008 report layout.

The marks can be enabled or disabled independently per company and are displayed in the document layout preview and rendered in the final PDF reports.


NICO-SOLUTIONS:
@CRogos 
@mt-software-de 
Hey, 
if you have some time, I’d appreciate it if you could review this PR.

smaddlsoft:
Functionally, LGTM (see first screenshot).

image

Minor point: You could improve the general tooltip (see second screenshot) to say "DIN fold and punch/hole marker" and you could harmonise using either punch, or hole, or punch/hole marker across the board.

image

smaddlsoft (approved):

NICO-SOLUTIONS:
@smaddlsoft 
thx for reviewing.
I changed the "Show Marker" label to a standalone <span> because it represents a heading for both marker settings and should not be associated with l10n_din5008_fold_marker (as it was before).

Previously, the label used for="l10n_din5008_fold_marker", which caused the field's help tooltip to be associated with the heading. The heading therefore no longer has a tooltip, and the help text is now defined directly on the respective fields instead.

I also harmonised the terminology to "punch/hole marker" across the module

smaddlsoft:
Better, though now you made the Show Marker entry in normal font, whereas all other are bold, sorry to be picky :)

image

NICO-SOLUTIONS:
> Better, though now you made the Show Marker entry in normal font, whereas all other are bold, sorry to be picky :)
> 
> 
@smaddlsoft 
It was actually a font-weight difference (500 vs. 400), not a bold vs. normal issue. 
Fixed now — thanks for being picky 😄

l10n_din5008_report_marks/models/__init__.py
```diff
@@ -0,0 +1,6 @@
+# Copyright (C) 2026-TODAY NICO SOLUTIONS - ENGINEERING & IT (<https://www.nico-solutions.de>)
+# @author Nils Coenen <nils.coenen@nico-solutions.de>
```
mt-software-de:
```suggestion
# Copyright 2026 NICO SOLUTIONS - ENGINEERING & IT (<https://www.nico-solutions.de>)
```
The most common way in OCA is `# Copyright 2026 <your string>`
Your copyright is already stated by git history and the copyright line, so `@author` isn't needed,
it just makes the headers of your python files bigger.

l10n_din5008_report_marks/models/__init__.py
```diff
@@ -0,0 +1,6 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from . import res_company
+from . import ir_action_report
```
mt-software-de:
```suggestion
from . import ir_actions_report
```

l10n_din5008_report_marks/models/ir_actions_report.py
mt-software-de, line 1: Did you tried to do it without any new any new dependencies, just by inheriting the template and using basic css?
This would keep the external dependencies list empty. 

I did it once to add an small watermark image. 
```
<template
    id="override_external_layout_standard"
    inherit_id="web.external_layout_standard"
    name="xx"
    priority="100"
    >
                <img
                    src="image path"
                    style="position: absolute; left: 12px; top: 650px; width: 50px;"
                />
```

l10n_din5008_report_marks/__manifest__.py
```diff
@@ -0,0 +1,29 @@
...
+    "name": "L10n DIN5008 Report Marks",
+    "summary": "This module adds configurable fold and punch/hole "
+    "marks to the DIN 5008 "
+    "external layout.",
+    "version": "19.0.1.0.0",
+    "license": "AGPL-3",
+    "author": "NICO SOLUTIONS - ENGINEERING & IT, Odoo Community Association (OCA)",
+    "maintainers": ["NICO-SOLUTIONS"],
+    "website": "https://github.com/OCA/l10n-germany",
+    "depends": ["web", "l10n_din5008"],
```
mt-software-de:
```suggestion
    "depends": ["l10n_din5008"],
```

mt-software-de (changes requested):
Thx for your contribution.
I would welcome it if it could be done without further dependencies.

NICO-SOLUTIONS:
@mt-software-de 
Hey, I’ll have a look at the suggestions later. I’d also suggest having a short discussion about the overall structure first, as there are some refactored modules coming up that might affect it.
image

l10n_din5008_report_marks/models/ir_actions_report.py
NICO-SOLUTIONS: Yes, I tried the CSS/QWeb approach without adding any dependencies. 
It's actually active for the preview, and the QWeb part works fine — the marks are present in the generated HTML and also show up in the document layout preview. The issue is the final PDF: as soon as the marks need to be placed in the 20 mm left margin, they disappear.

I tried a few different CSS approaches, but couldn't get them into the actual paper margin. That's why I took the Python approach for this. Not to add a dependency unnecessarily, but because I couldn't get the required result with QWeb/CSS alone.

NICO-SOLUTIONS: @mt-software-de 
I think I found another way using _build_wkhtmltopdf_args. I need to do some testing first...

NICO-SOLUTIONS: @mt-software-de 
I have found a different solution...
The approach using `_build_wkhtmltopdf_args` caused too many side effects, particularly when extending the page/content area to place the marks in the left margin area. Instead, I now add the marks directly to the PDF as an overlay after rendering with wkhtmltopdf, using the helper functions provided by `odoo.tools.pdf`.

This means that no additional Python dependencies are required, so the entire external_dependencies section has been removed from the manifest.

PR #258, opened 2026-09-15
[19.0][ADD] l10n_din5008_report_company_sender_address: new module

NICO-SOLUTIONS:
This module extends the DIN 5008 external report layout with configurable options for the company sender address.

The standard DIN 5008 sender address may not provide sufficient flexibility for different company layouts and addressing requirements. This module allows the company sender address to be customized while preserving the DIN 5008 report layout.

Features:
- Configure the country indicator displayed in the company sender address.
- Choose between the country code and country name, or hide the country indicator.
- Configure the separator used between individual address components.
- Choose between a pipe (`|`) and a dot (`•`) as the address separator.
- Configure the font size scale factor for the company sender address.
- Restrict the sender address font scale factor to a valid range of 0.5 to 2.0.
- Apply the configured settings to reports using the DIN 5008 external report layout.

The settings are available in the company document layout configuration and apply to reports using the DIN 5008 external report layout.


NICO-SOLUTIONS:
@mt-software-de 
Hi,
I’m currently refactoring my modules, including l10n_din5008_report_company_sender_address.

I noticed that it conflicts with l10n_din5008_company_header_country_invisible because both modify the same part of the DIN 5008 sender address. Would you prefer that I make my module depend on yours to avoid the conflict? I’d rather avoid that if possible, so I’m wondering if you see a better way to make both modules work independently.

What do you think?

mt-software-de:
> @mt-software-de Hi, I’m currently refactoring my modules, including l10n_din5008_report_company_sender_address.
> 
> I noticed that it conflicts with l10n_din5008_company_header_country_invisible because both modify the same part of the DIN 5008 sender address. Would you prefer that I make my module depend on yours to avoid the conflict? I’d rather avoid that if possible, so I’m wondering if you see a better way to make both modules work independently.
> 
> What do you think?

Thx. 
I think we should merge the add-ons into an l10n_din5008_base_report.
Also with your changes from https://github.com/OCA/l10n-germany/pull/255 (if we get rid of the new dependencies).

The add-on could later serve as a base for the other DIN 5008 report add-ons, making them configurable as well and allowing them to be displayed in the same location within the view. Like `l10n_din5008_move_name` and `l10n_din5008_sale_move_name`

